PR #23, opened 2019-05-16
Add VSCode as IDE

8sd:
Add `.vscode` folder to gitiginore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea/
+.vscode/
 __pycache__/
 *.pkl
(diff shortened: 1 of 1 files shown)


rralf:
Hi,

this is a local artefact that only appears for users of Visual Studio (I guess). A better solution is to add this directory to your global gitignore, e.g., by using:

git config --global core.excludesfile ~/.gitignore

But: Consequently, I'm willing to accept a patch that removes .idea from .gitignore.

Thanks

PR #24, opened 2019-05-16
Remove `.idea`

8sd:
As discussed, `.idea` is removed from the local `.gitignore` file and can be added by developers needing it to their global `.gitignore` file.
This can be achieved by executing `git config --global core.excludesfile ~/.gitignore`.


rralf:
Thanks, merged to next.

PR #26, opened 2019-05-25
Fix spelling consistency

8sd:


bulwahn:
If we are considering capitalization, we should consider grammar:
Then it would be `Interactively accepted` and `Interactively declined` .

rralf:
Is this already a v2 of the patch? Looks good to me.

rralf:
Will squash this to one patch and merge. Thanks!

PR #37, opened 2019-12-03
Add name of author to dump

8sd:
Add name from `c.mail_from[0]` to the csv dump of the patches.

### DIFF
--- a/bin/pasta_evaluate_patches.py
+++ b/bin/pasta_evaluate_patches.py
@@ -220,7 +220,7 @@ def get_kv_rc(linux_version):
 
 def dump_characteristics(repo, characteristics, ignored, relevant, filename):
     with open(filename, 'w') as csv_file:
-        csv_fields = ['id', 'from', 'list', 'list_matches_patch', 'kv', 'rc',
+        csv_fields = ['id', 'from', 'from_name', 'list', 'list_matches_patch', 'kv', 'rc',
                       'ignored', 'time']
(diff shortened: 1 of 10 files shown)


rralf:
Could you please explain why you want to add this field? I can't see any users of that field.

8sd:
I need the name for the analyses if authors are discriminated.
The tool I use to guess sex/country is based on the name. 
Having this data would significantly simplify this analysis.

bin/pasta_evaluate_patches.py
```diff
@@ -240,6 +240,7 @@ def dump_characteristics(repo, characteristics, ignored, relevant, filename):
 
                 row = {'id': message_id,
                        'from': mail_from,
+                       'from_name': c.mail_from[0],
```
rralf, line 243: Guess that should be refactored to avoid confusion.
from -> from_email.

rralf: And have a look at line 231. We can simplify it to sth. like:

from_email, from_name = c.mail_from.

rralf:
I won't accept it unless we have upstream PaStA users of this value (e.g., in an R evaluation script). Adding superfluous values makes maintainence hard: At the moment, I can tell you with one look at the list which variables are used in any analysis, and which aren't. If we add optional values, this is not the case any longer. However, I'm willing to accept if we have users for the variable (with respect to the comments in the patch).

PR #50, opened 2020-04-04
pypasta: config: change default filename for cluster output

rsarky:
The default filename for both patch stack and mbox mode is same now

Signed-off-by: Rohit Sarkar <rohitsarkar5398@gmail.com>

### DIFF
--- a/pypasta/Config.py
+++ b/pypasta/Config.py
@@ -151,11 +151,7 @@ class Config:
         # analysis results
         self.d_false_positives = path('FALSE_POSTITIVES')
 
-        if self._mode == Config.Mode.MBOX:
-            cluster_dst = 'MBOX_RESULT'
-        else:
-            cluster_dst = 'PASTA_RESULT'
-        self.f_clustering = path(cluster_dst)
+        self.f_clustering = path('PATCH_GROUPS')
 
         self.f_commit_description = path('COMMIT_DESCRIPTION')
 


pypasta/Config.py
```diff
@@ -151,10 +151,7 @@ def path(name):
         # analysis results
         self.d_false_positives = path('FALSE_POSTITIVES')
 
-        if self._mode == Config.Mode.MBOX:
-            cluster_dst = 'MBOX_RESULT'
-        else:
-            cluster_dst = 'PASTA_RESULT'
+        cluster_dst = 'PATCH_GROUPS'
```
rralf: One last thing: You can use 'PATCH_GROUPS' directly below. There's no need for outsourcing it in a variable.

pypasta/Config.py
```diff
@@ -151,10 +151,7 @@ def path(name):
...
         self.f_clustering = path(cluster_dst)
```
rralf: Just use path('PATCH_GROUPS')

rralf (changes requested):
Almost :-) Just one last simplification, then I'll integrate it. Thanks!

pypasta/Config.py
rsarky: yup makes sense

rralf:
Thanks, applied!

rralf:
Ah crap, that was a bit too fast: Could you please forward the resources submodule?

rsarky:
Sure! Will do once the PR is merged there.

rralf:
It is merged.

rsarky:
[This](https://github.com/lfd/PaStA-resources/pull/2) isnt merged yet

PR #52, opened 2020-04-07
Script for maintainers

basakerdamar:
This script allows to parse MAINTAINERS file using LinuxMaintainers class, reports subsystems or maintainers optionally filtered by a given list of files together with LoC and byte counts for the files.
A command like: pasta maintainers show [--smallstat] [--outfile OUTFILE] [--file FILE] [--filter FILTER] would provide the option to filter by the contents of FILTER, allow to use another MAINTAINERS file FILE other than the one on the repository, and see a smaller view with less fields in the report.

### DIFF
--- a/bin/pasta_maintainers_stats.py
+++ b/bin/pasta_maintainers_stats.py
@@ -1,0 +1,275 @@
+"""
+PaStA - Patch Stack Analysis
+Copyright (c) Bayerische Motoren Werke Aktiengesellschaft (BMW AG), 2020
+Copyright (c) OTH Regensburg, 2019-2020
+Authors:
+  Basak Erdamar <basakerdamar@gmail.com>
+This work is licensed under the terms of the GNU GPL, version 2. See
+the COPYING file in the top-level directory.
+"""
+
+import os
+import pygit2
+import sys
+
+from argparse import ArgumentParser
+from collections import defaultdict
+from csv import writer
+from logging import getLogger
+from multiprocessing import Pool, cpu_count
+
+sys.path.insert(0,os.path.abspath(os.path.join(os.path.dirname(__file__),'..')))
+
+from pypasta.LinuxMaintainers import load_maintainer
+from pypasta.Util import file_to_string
+
+log = getLogger(__name__[-15:])
+
+# We need this global variable, as pygit2 Commit.tree objects are not pickleable
+_tmp_tree = None
+
+
+
+def walk_commit_tree(tree):
+    results = set()
+
+    for entry in tree:
+        if type(entry) == pygit2.Blob:
+            results.add(entry.name)
+        elif type(entry) == pygit2.Tree:
+            results |= {os.path.join(entry.name, item)
+                        for item in walk_commit_tree(entry)}
+        else:
+            raise TypeError('Unknown type: %s' % type(entry))
+
+    return results
+
+
+def get_tree(repo, revision):
+    if revision == 'HEAD':
+        return repo.revparse_single(repo.head.name).tree
+
+    if revision.startswith('v'):
+        revision = repo.lookup_reference('refs/tags/%s' % revision).target
+    commit_hash = repo[revision].target
+    tree = repo[commit_hash].tree
+
+    return tree
+
+
+def decreasing_order(dictionary):
+    return sorted(dictionary.items(), key=lambda item:item[1].loc, reverse=True)
+
+
+def get_file_size(filename):
+    blob = _tmp_tree[filename]
+    lines = blob.data.count(b'\n')
+    size = blob.size
+
+    return filename, (lines, size)
+
+
+def pretty_name(maintainer):
+    return maintainer[0]+' <'+maintainer[1]+'>'
+
+
+def status(all_maintainers, section_name):
+    if any(all_maintainers.subsystems[section_name].status):
+        return all_maintainers.subsystems[section_name].status[0].value
+    else:
+        return ''
+
+
+class Counter:
+    def __init__(self, loc=0, byte=0, filter=None):
+        self.loc = loc
+        self.byte = byte
+        self.loc_filt = 0
+        self.byte_filt = 0
+        if filter:
+            self.loc_filt = loc
+            self.byte_filt = byte
+
+    def increase(self, loc, byte, filter):
+        self.loc += loc
+        self.byte += byte
+        if filter:
+            self.loc_filt += loc
+            self.byte_filt += byte
+
+    def display_in_filter(self):
+        # whether to show this item in filtered view or not
+        if self.byte_filt > 0:
+            return True
+        return False
+
+
+def get_counts(all_maintainers, file_sizes, filter_by_files, group_by):
+    counts = defaultdict(Counter)
+    if  group_by == 'sections':
+        filename_excluded = defaultdict(list)
+        filenames_filt = defaultdict(list)
+    else:
+        maintainer_to_section = defaultdict(set)
+    for filename, (loc, byte) in file_sizes.items():
+        sections = all_maintainers.get_subsystems_by_file(filename)
+        sections -= {'THE REST'}
+        for section in sections:
+            if group_by == 'sections':
+                counts[section].increase(loc, byte, filename in filter_by_files)
+                if filename in filter_by_files:
+                    filenames_filt[section].append(filename)
+                else:
+                    filename_excluded[section].append(filename)
+            else:
+                _, maintainers, _ = all_maintainers.get_maintainers(section)
+                for maintainer in maintainers:
+                    name = pretty_name(maintainer)
+                    maintainer_to_section[name].add(section)
+                    counts[name].increase(loc, byte, filename in filter_by_files)
+
+    if len(filter_by_files) > 0:
+        # If filter is given, only return the relevant lines:
+        counts = {name:item for name,item in counts.items()
+                    if item.display_in_filter()}
+    if group_by == 'sections':
+        return decreasing_order(counts), filename_excluded, filenames_filt
+    else:
+        return decreasing_order(counts), maintainer_to_section
+
+
+def get_sections_by_files(all_maintainers, filenames):
+    results = []
+    for filename in filenames:
+        sections = all_maintainers.get_subsystems_by_file(filename)
+        sections -= {'THE REST'}
+        # To get a sense of how large the section is and order by it, look at
+        # how many different patterns it includes:
+        # all_maintainers.subsystems[section].files
+        sections = sorted(sections, 
+         key=lambda section: len(all_maintainers.subsystems[section].files))
+        results.append([filename, sections])
+    return results
+
+
+def maintainers_stats(config, sub, argv):
+    parser = ArgumentParser(description='Display file sizes grouped by '
+                                            'maintainers or sections')
+    parser.add_argument('maintainers_stats', nargs=1, type=bool)
+    parser.add_argument('--smallstat', action='store_true', help='Simple view')
+    parser.add_argument('--bytes', action='store_true', help='Show byte counts')
+    parser.add_argument('--outfile', type=str, help='Output to a csv file')
+    parser.add_argument('--filter', type=str, help='Filter by file list: '
+                        'enter the file name for the file containing the list '
+                        'of files to filter by.')
+    parser.add_argument('--group-by', type=str, default='sections',
+                        help='files option '
+                        'Displays files with corresponding sections\n'
+                        'sections option groups files by sections, displays '
+                        'sections ordered by lines of code they include\n'
+                        'maintainers option groups files by sections, displays '
+                        'each maintainer ordered by lines of code they are '
+                        'responsible for\n'
+                        'Default is sections if no value is specified')
+    parser.add_argument('--filesize', action='store_true', 
+                                        help='Display file sizes in file mode')
+    parser.add_argument('--revision', type=str, help='Specify a commit hash or '
+                        'a version name for a Linux repo')
+
+    args = parser.parse_args()
+    repo = config.repo
+
+    kernel_revision = 'HEAD'
+    filter_by_files = list()
+    if args.filter:
+        filter_by_files = file_to_string(args.filter).splitlines()
+        # The first line of the file must contain a valid kernel version,
+        # tag or commit hash
+        kernel_revision = filter_by_files.pop(0)
+
+    # arguments may override the kernel revision, if it is not already set
+    if args.revision:
+        kernel_revision = args.revision
+    log.info('Working on kernel revision %s' % kernel_revision)
+
+    all_maintainers = load_maintainer(repo, kernel_revision)
+    tree = get_tree(repo.repo, kernel_revision)
+    all_filenames = walk_commit_tree(tree)
+
+    global _tmp_tree
+    _tmp_tree = tree
+    processes = int(cpu_count())
+    p = Pool(processes=processes, maxtasksperchild=1)
+    result = p.map(get_file_size, all_filenames)
+    p.close()
+    p.join()
+    _tmp_tree = None
+
+    file_sizes = dict(result)
+
+    if  args.group_by == 'sections':# or not args.group_by:
+        title = ['Sections','LoC']
+        results, irrelevant_dirs, relevant_dirs = \
+         get_counts(all_maintainers, file_sizes, filter_by_files, args.group_by)
+    elif args.group_by == 'maintainers':
+        title = ['Maintainers', 'LoC']
+        results, maintainer_to_section = \
+         get_counts(all_maintainers, file_sizes, filter_by_files, args.group_by)
+    elif args.group_by == 'files':
+        title = ['File name', 'Sections of file']
+        victims = filter_by_files if len(filter_by_files) else all_filenames
+        results = get_sections_by_files(all_maintainers, victims)
+    else:
+        parser.print_help()
+        sys.exit()
+
+    fields = dict()
+    fields['Maintainers'] = fields['Sections'] = fields['File name'] =\
+                                                         lambda item : item[0]
+    fields['LoC in list'] = lambda item : item[1].loc_filt
+    fields['Total LoC'] = fields['LoC'] = lambda item : item[1].loc
+    fields['Byte count'] = lambda item : item[1].byte
+    fields['In list(%)']=lambda item:round((item[1].loc_filt/item[1].loc)*100,2)
+    fields['Status'] = lambda item : status(all_maintainers, item[0])
+    fields['Sections of maintainer'] = \
+                         lambda item:', '.join(maintainer_to_section[item[0]])
+    fields['Irrelevant files'] = lambda item: ','.join(irrelevant_dirs[item[0]])
+    fields['Relevant files'] = lambda item: ','.join(relevant_dirs[item[0]])
+    fields['Sections of file'] = lambda item : ','.join(item[1])
+    fields['LoC of file'] = lambda item : file_sizes[item[0]][0]
+
+    if args.bytes:
+        title += ['Byte count']
+    if not args.smallstat:
+        if args.filter and not args.group_by == 'files':
+            title += ['LoC in list', 'In list(%)']
+            if  args.group_by == 'sections':
+                title += ['Relevant files', 'Irrelevant files']
+        if  args.group_by == 'sections':
+            title.insert(1, 'Status')
+        elif args.group_by =='maintainers':
+            title += ['Sections of maintainer']
+        elif  args.group_by == 'files' and args.filesize:
+            title.insert(1, 'LoC of file')
+
+
+    results = [[fields[field](item) for field in title] for item in results]
+    results.insert(0, title)
+    if args.outfile:
+        with open(args.outfile, 'w+') as csv_file:
+            csv_writer = writer(csv_file)
+            csv_writer.writerows(results)
+    else:
+        width = len(max(list(zip(*results))[0], key = len))
+        _, columns = os.popen('stty size', 'r').read().split()
+        columns = int(columns)
+        for line in results:
+            line_s = str(line[0]).ljust(width+4)
+            for item in line[1:]:
+                line_s += str(item).ljust(15)
+            if len(line_s) > columns:
+                print(line_s[0:columns-3]+"...")
+            else:
+                print(line_s)
+
+    return 0
--- a/bin/pasta_maintainers_stats.py
+++ b/bin/pasta_maintainers_stats.py
@@ -1,9 +1,13 @@
 """
 PaStA - Patch Stack Analysis
+
 Copyright (c) Bayerische Motoren Werke Aktiengesellschaft (BMW AG), 2020
-Copyright (c) OTH Regensburg, 2019-2020
+Copyright (c) OTH Regensburg, 2020
+
 Authors:
   Basak Erdamar <basakerdamar@gmail.com>
+  Ralf Ramsauer <ralf.ramsauer@oth-regensburg.de
+
 This work is licensed under the terms of the GNU GPL, version 2. See
 the COPYING file in the top-level directory.
 """
@@ -13,7 +17,7 @@ import pygit2
 import sys
 
 from argparse import ArgumentParser
-from collections import defaultdict
+from collections import defaultdict, Counter
 from csv import writer
 from logging import getLogger
 from multiprocessing import Pool, cpu_count
@@ -27,7 +31,7 @@ log = getLogger(__name__[-15:])
 
 # We need this global variable, as pygit2 Commit.tree objects are not pickleable
 _tmp_tree = None
-
+_all_maintainers = None
 
 
 def walk_commit_tree(tree):
@@ -57,16 +61,14 @@ def get_tree(repo, revision):
     return tree
 
 
-def decreasing_order(dictionary):
-    return sorted(dictionary.items(), key=lambda item:item[1].loc, reverse=True)
-
-
-def get_file_size(filename):
+def get_file_map(filename):
     blob = _tmp_tree[filename]
     lines = blob.data.count(b'\n')
     size = blob.size
 
-    return filename, (lines, size)
+    sections = _all_maintainers.get_subsystems_by_file(filename)
+
+    return filename, (lines, size, sections)
 
 
 def pretty_name(maintainer):
@@ -80,76 +82,18 @@ def status(all_maintainers, section_name):
         return ''
 
 
-class Counter:
-    def __init__(self, loc=0, byte=0, filter=None):
-        self.loc = loc
-        self.byte = byte
-        self.loc_filt = 0
-        self.byte_filt = 0
-        if filter:
-            self.loc_filt = loc
-            self.byte_filt = byte
+def dump_csv(title, data, filename):
+    if filename:
+        # implement csv writing stuff...
+        return
 
-    def increase(self, loc, byte, filter):
-        self.loc += loc
-        self.byte += byte
-        if filter:
-            self.loc_filt += loc
-            self.byte_filt += byte
+    # El Cheapo pretty printer.
+    headers = '\t\t'.join([header[0] for header in title])
+    formats = '\t\t'.join([header[1] for header in title])
 
-    def display_in_filter(self):
-        # whether to show this item in filtered view or not
-        if self.byte_filt > 0:
-            return True
-        return False
-
-
-def get_counts(all_maintainers, file_sizes, filter_by_files, group_by):
-    counts = defaultdict(Counter)
-    if  group_by == 'sections':
-        filename_excluded = defaultdict(list)
-        filenames_filt = defaultdict(list)
-    else:
-        maintainer_to_section = defaultdict(set)
-    for filename, (loc, byte) in file_sizes.items():
-        sections = all_maintainers.get_subsystems_by_file(filename)
-        sections -= {'THE REST'}
-        for section in sections:
-            if group_by == 'sections':
-                counts[section].increase(loc, byte, filename in filter_by_files)
-                if filename in filter_by_files:
-                    filenames_filt[section].append(filename)
-                else:
-                    filename_excluded[section].append(filename)
-            else:
-                _, maintainers, _ = all_maintainers.get_maintainers(section)
-                for maintainer in maintainers:
-                    name = pretty_name(maintainer)
-                    maintainer_to_section[name].add(section)
-                    counts[name].increase(loc, byte, filename in filter_by_files)
-
-    if len(filter_by_files) > 0:
-        # If filter is given, only return the relevant lines:
-        counts = {name:item for name,item in counts.items()
-                    if item.display_in_filter()}
-    if group_by == 'sections':
-        return decreasing_order(counts), filename_excluded, filenames_filt
-    else:
-        return decreasing_order(counts), maintainer_to_section
-
-
-def get_sections_by_files(all_maintainers, filenames):
-    results = []
-    for filename in filenames:
-        sections = all_maintainers.get_subsystems_by_file(filename)
-        sections -= {'THE REST'}
-        # To get a sense of how large the section is and order by it, look at
-        # how many different patterns it includes:
-        # all_maintainers.subsystems[section].files
-        sections = sorted(sections, 
-         key=lambda section: len(all_maintainers.subsystems[section].files))
-        results.append([filename, sections])
-    return results
+    print(headers)
+    for entry in data:
+        print(formats % entry)
 
 
 def maintainers_stats(config, sub, argv):
@@ -196,80 +140,122 @@ def maintainers_stats(config, sub, argv):
     tree = get_tree(repo.repo, kernel_revision)
     all_filenames = walk_commit_tree(tree)
 
+    if args.group_by == 'files':
+        filenames = filter_by_files if len(filter_by_files) else all_filenames
+
+        title = [('%30s' % 'Filename', '%30.30s'),
+                 ('Sections of file', '%s')]
+        results = []
+        for filename in filenames:
+            sections = all_maintainers.get_subsystems_by_file(filename)
+            sections -= {'THE REST'}
+            results.append((filename, sections))
+
+        dump_csv(title, results, args.outfile)
+        return
+
+    # We end up here in case of args.group_by in {'sections', 'maintainers'}
     global _tmp_tree
+    global _all_maintainers
+
     _tmp_tree = tree
+    _all_maintainers = all_maintainers
     processes = int(cpu_count())
     p = Pool(processes=processes, maxtasksperchild=1)
-    result = p.map(get_file_size, all_filenames)
+    file_map = p.map(get_file_map, all_filenames)
     p.close()
     p.join()
+
     _tmp_tree = None
+    _all_maintainers = None
 
-    file_sizes = dict(result)
+    file_map = dict(file_map)
 
-    if  args.group_by == 'sections':# or not args.group_by:
-        title = ['Sections','LoC']
-        results, irrelevant_dirs, relevant_dirs = \
-         get_counts(all_maintainers, file_sizes, filter_by_files, args.group_by)
-    elif args.group_by == 'maintainers':
-        title = ['Maintainers', 'LoC']
-        results, maintainer_to_section = \
-         get_counts(all_maintainers, file_sizes, filter_by_files, args.group_by)
-    elif args.group_by == 'files':
-        title = ['File name', 'Sections of file']
-        victims = filter_by_files if len(filter_by_files) else all_filenames
-        results = get_sections_by_files(all_maintainers, victims)
+    # An object is the kind of the analysis, and reflects the target. A target
+    # can either be a section or the maintainer.
+
+    # Maps Section/Maintainer to a Counter
+    object_stats = defaultdict(Counter)
+    # Maps Section/Maintainer to a set of files
+    object_files = defaultdict(set)
+
+    relevant = defaultdict(Counter)
+
+    #############################################
+    # Routines that handle differences between
+    # maintainer / section grouping
+    #############################################
+    # Routines that are chosen in case of --group-by maintainers
+    def _evaluator_maintainers(file, lines, size, section):
+        _, mtrs, _ = all_maintainers.get_maintainers(section)
+        for mtr in mtrs:
+            object_files[mtr].add(file)
+            object_stats[mtr].update(lines=lines, size=size)
+
+    def _filter_sections(section, lines, size):
+        relevant[section].update(lines=lines, size=size)
+
+    # Routines that are chosen in case of --group-by sections
+    def _evaluator_sections(file, lines, size, section):
+        object_files[section].add(file)
+        object_stats[section].update(lines=lines, size=size)
+
+    def _filter_maintainers(section, lines, size):
+        _, mtrs, _ = all_maintainers.get_maintainers(section)
+        for mtr in mtrs:
+            relevant[mtr].update(lines=lines, size=size)
+    #############################################
+
+    # Choose the right routines
+    _title, _evaluator, _filter = {
+        'maintainers': ('Maintainers',
+                        _evaluator_maintainers,
+                        _filter_maintainers),
+
+        'sections': ('Sections',
+                     _evaluator_sections,
+                     _filter_sections),
+    }[args.group_by]
+
+    # First of all, fill object_{stats, file}
+    for file, (lines, size, sections) in file_map.items():
+        for section in sections:
+            _evaluator(file, lines, size, section)
+
+    # Do we have to respect any filters?
+    for file in filter_by_files:
+        lines, size, sections = file_map[file]
+        for section in sections:
+            _filter(section, lines, size)
+
+    # Fill the first two fields of the title
+    title = [('%30s' % _title, '%30.30s'),
+             ('Sum Lines', '%9u')]
+
+    result = list()
+    if len(relevant):
+        title += [('Lines in filter',  '%15u'),
+                  ('Lines percentage', '%16.2f'),
+                  ('Relevant files',   '%s')]
+
+        for object, counter in relevant.items():
+            object_stat = object_stats[object]
+            lines_percentage = counter['lines'] / object_stat['lines']
+
+            result.append((object,
+                           object_stat['lines'],
+                           counter['lines'],
+                           lines_percentage,
+                           {filename for filename in object_files[object] if
+                            filename in filter_by_files}))
+
+        # sort by lines percentage
+        result.sort(key=lambda x: x[3])
     else:
-        parser.print_help()
-        sys.exit()
+        for object, counter in object_stats.items():
+            result.append((object, counter['lines']))
 
-    fields = dict()
-    fields['Maintainers'] = fields['Sections'] = fields['File name'] =\
-                                                         lambda item : item[0]
-    fields['LoC in list'] = lambda item : item[1].loc_filt
-    fields['Total LoC'] = fields['LoC'] = lambda item : item[1].loc
-    fields['Byte count'] = lambda item : item[1].byte
-    fields['In list(%)']=lambda item:round((item[1].loc_filt/item[1].loc)*100,2)
-    fields['Status'] = lambda item : status(all_maintainers, item[0])
-    fields['Sections of maintainer'] = \
-                         lambda item:', '.join(maintainer_to_section[item[0]])
-    fields['Irrelevant files'] = lambda item: ','.join(irrelevant_dirs[item[0]])
-    fields['Relevant files'] = lambda item: ','.join(relevant_dirs[item[0]])
-    fields['Sections of file'] = lambda item : ','.join(item[1])
-    fields['LoC of file'] = lambda item : file_sizes[item[0]][0]
+        # sort by sum lines
+        result.sort(key=lambda x: x[1])
 
-    if args.bytes:
-        title += ['Byte count']
-    if not args.smallstat:
-        if args.filter and not args.group_by == 'files':
-            title += ['LoC in list', 'In list(%)']
-            if  args.group_by == 'sections':
-                title += ['Relevant files', 'Irrelevant files']
-        if  args.group_by == 'sections':
-            title.insert(1, 'Status')
-        elif args.group_by =='maintainers':
-            title += ['Sections of maintainer']
-        elif  args.group_by == 'files' and args.filesize:
-            title.insert(1, 'LoC of file')
-
-
-    results = [[fields[field](item) for field in title] for item in results]
-    results.insert(0, title)
-    if args.outfile:
-        with open(args.outfile, 'w+') as csv_file:
-            csv_writer = writer(csv_file)
-            csv_writer.writerows(results)
-    else:
-        width = len(max(list(zip(*results))[0], key = len))
-        _, columns = os.popen('stty size', 'r').read().split()
-        columns = int(columns)
-        for line in results:
-            line_s = str(line[0]).ljust(width+4)
-            for item in line[1:]:
-                line_s += str(item).ljust(15)
-            if len(line_s) > columns:
-                print(line_s[0:columns-3]+"...")
-            else:
-                print(line_s)
-
-    return 0
+    dump_csv(title, result, args.outfile)
--- a/bin/pasta_maintainers_stats.py
+++ b/bin/pasta_maintainers_stats.py
@@ -75,7 +75,7 @@ def pretty_name(maintainer):
     return maintainer[0]+' <'+maintainer[1]+'>'
 
 
-def status(all_maintainers, section_name):
+def get_status(all_maintainers, section_name):
     if any(all_maintainers.subsystems[section_name].status):
         return all_maintainers.subsystems[section_name].status[0].value
     else:
@@ -83,8 +83,12 @@ def status(all_maintainers, section_name):
 
 
 def dump_csv(title, data, filename):
+
     if filename:
-        # implement csv writing stuff...
+        with open(filename, 'w+') as csv_file:
+            csv_writer = writer(csv_file)
+            csv_writer.writerow([header[0] for header in title])
+            csv_writer.writerows(data)
         return
 
     # El Cheapo pretty printer.
@@ -107,16 +111,15 @@ def maintainers_stats(config, sub, argv):
                         'enter the file name for the file containing the list '
                         'of files to filter by.')
     parser.add_argument('--group-by', type=str, default='sections',
+                        choices={'maintainers', 'sections', 'files'},
                         help='files option '
-                        'Displays files with corresponding sections\n'
+                        'displays files with corresponding sections. '
                         'sections option groups files by sections, displays '
-                        'sections ordered by lines of code they include\n'
+                        'sections ordered by lines of code they include. '
                         'maintainers option groups files by sections, displays '
                         'each maintainer ordered by lines of code they are '
-                        'responsible for\n'
+                        'responsible for. '
                         'Default is sections if no value is specified')
-    parser.add_argument('--filesize', action='store_true', 
-                                        help='Display file sizes in file mode')
     parser.add_argument('--revision', type=str, help='Specify a commit hash or '
                         'a version name for a Linux repo')
 
@@ -143,13 +146,13 @@ def maintainers_stats(config, sub, argv):
     if args.group_by == 'files':
         filenames = filter_by_files if len(filter_by_files) else all_filenames
 
-        title = [('%30s' % 'Filename', '%30.30s'),
-                 ('Sections of file', '%s')]
+        title = [('%-30s' % 'Filename', '%-30.30s'),
+                 ('Sections of file', '%-s')]
         results = []
         for filename in filenames:
             sections = all_maintainers.get_subsystems_by_file(filename)
             sections -= {'THE REST'}
-            results.append((filename, sections))
+            results.append((filename, ', '.join(sections)))
 
         dump_csv(title, results, args.outfile)
         return
@@ -229,31 +232,62 @@ def maintainers_stats(config, sub, argv):
             _filter(section, lines, size)
 
     # Fill the first two fields of the title
-    title = [('%30s' % _title, '%30.30s'),
-             ('Sum Lines', '%9u')]
+    title = [('%-30s' % _title, '%-30.30s'),
+             ('Sum Lines', '%-9u')]
+
+    if args.bytes:
+        title += [('File size',   '%-15u')]
+
+    if args.group_by == 'sections':
+        title += [('Status',   '%-10s')]
 
     result = list()
+
     if len(relevant):
-        title += [('Lines in filter',  '%15u'),
-                  ('Lines percentage', '%16.2f'),
-                  ('Relevant files',   '%s')]
+        title += [('Lines in filter',  '%-15u'),
+                  ('Lines percentage', '%-16.2f'),
+                  ('Relevant files',   '%-s')]
+
 
         for object, counter in relevant.items():
             object_stat = object_stats[object]
             lines_percentage = counter['lines'] / object_stat['lines']
 
-            result.append((object,
-                           object_stat['lines'],
-                           counter['lines'],
-                           lines_percentage,
-                           {filename for filename in object_files[object] if
-                            filename in filter_by_files}))
+            item = (object if not args.group_by == 'maintainers' 
+                                            else pretty_name(object),
+                    object_stat['lines'])
+            
+            if args.bytes:
+                item += (object_stat['size'],)
+
+            if args.group_by == 'sections':
+                item += (get_status(all_maintainers, object), )
+
+            item += (
+                     counter['lines'],
+                     lines_percentage,
+                     ', '.join({filename for filename in object_files[object] 
+                                               if filename in filter_by_files}))
+
+
+            result.append(item)
 
         # sort by lines percentage
         result.sort(key=lambda x: x[3])
     else:
         for object, counter in object_stats.items():
-            result.append((object, counter['lines']))
+
+            item = (object if not args.group_by == 'maintainers' 
+                                            else pretty_name(object),
+                    counter['lines'])
+
+            if args.bytes:
+                item += (counter['size'],)
+
+            if args.group_by == 'sections':
+                item += (get_status(all_maintainers, object), )
+
+            result.append(item)
 
         # sort by sum lines
         result.sort(key=lambda x: x[1])
--- a/pasta
+++ b/pasta
@@ -31,8 +31,9 @@ from bin.pasta_check_connectivity import check_connectivity
 from bin.pasta_check_mbox import check_mbox
 from bin.pasta_compare import compare
 from bin.pasta_compare_clusters import compare_clusters
-from bin.pasta_prepare_evaluation import prepare_evaluation
+from bin.pasta_maintainers_stats import maintainers_stats
 from bin.pasta_optimise_cluster import optimise_cluster
+from bin.pasta_prepare_evaluation import prepare_evaluation
 from bin.pasta_rate import rate
 from bin.pasta_ripup import ripup
 from bin.pasta_show_cluster import show_cluster
@@ -83,6 +84,7 @@ def usage(me, exit_code=errno.EINVAL):
           '  ripup\n'
           '  upstream_history\n'
           '  web\n'
+          '  maintainers_stats\n'
           '\n'
           'If -c is not provided, PaStA will choose ./config as config file\n'
           '\n'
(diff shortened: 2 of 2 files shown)


rralf:
I'm missing a description and the context -- what are you trying to achieve? Thanks!

bulwahn:
@rralf, let Anmol and me first go through a code review with @basakerdamar; when we are happy with the code, you can do a second pass and give some more hints on things we might have overlooked.

rralf:
Alright!

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
+'''
+This work is licensed under the terms of the GNU GPL, version 2.  See
+the COPYING file in the top-level directory.
+'''
```
bulwahn: Add a copyright header and author as in https://github.com/lfd/PaStA/blob/master/bin/pasta_evaluate_patches.py

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+from pypasta.LinuxMaintainers import LinuxMaintainers
+from pypasta.Util import file_to_string
+
+log = getLogger(__name__[-15:])
+
+def flatten(nested):
+    return [item for sublist in nested for item in sublist]
+
+def shorten(string_list):
+    lenght = len(string_list)
```
bulwahn: typo it is length, not lenght.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+
+from pypasta.LinuxMaintainers import LinuxMaintainers
+from pypasta.Util import file_to_string
+
+log = getLogger(__name__[-15:])
+
+def flatten(nested):
+    return [item for sublist in nested for item in sublist]
+
+def shorten(string_list):
```
bulwahn: Function is not used anymore.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+
+def file_len(filename):
+    f = file_to_string(filename)
+    try:
+        return [f.count('\n'), len(f)]
+    # Empty file:
+    except:
+        return [0, 0]
+
+all_filenames = [os.path.join(r, item)[23:] for r, _, f in os.walk('./resources/linux/repo') for item in f]
```
bulwahn: You should not hard-code the subpath, you should take that from the pasta config. Of course, it only makes sense if the config is a linux config anyway.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+    file_lengths[filename] = file_len('./resources/linux/repo/' + str.strip(filename))
+
+def get_entry_counts(all_maintainers, filenames = all_filenames):
+    subsystem_dirs = dict()
+    counts_by_entry = dict()
+    for filename in filenames:
+        subsystems = all_maintainers.get_subsystems_by_file(filename)
+        for subsystem in subsystems:
+            subsystem_dirs[subsystem] = subsystem_dirs.get(subsystem, set())
+            subsystem_dirs[subsystem].add(''.join(re.findall('[a-zA-Z0-9_]+/', filename)))
```
bulwahn: what is this regular expression for?

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+'''
+
+import sys
+import os
+
+from csv import writer
+from logging import getLogger
+from collections import OrderedDict
+from argparse import ArgumentParser
+import numpy as np
```
bulwahn: shortening it to np is really overdoing it, I think.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+
+def get_maintainer_counts(all_maintainers, filenames = all_filenames):
+    maintainers_subs = dict()
+    counts_by_maintainer = dict()
+    for filename in filenames:
+        subsystems = all_maintainers.get_subsystems_by_file(filename)
+        for subsystem in subsystems:
+            maintainers = all_maintainers.get_maintainers(subsystem)
+            for maintainer in flatten(maintainers):
+                if not (type(maintainer) == str or maintainer[0] == '' or maintainer[1] == ''):
```
bulwahn: If that does not fit, you should at least about it in some verbose mode, right?

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+                    maintainers_subs[maintainer[0]+' <'+maintainer[1]+'>'] = maintainers_subs.get(maintainer[0]+' <'+maintainer[1]+'>', set())
```
bulwahn: maintainer[0]+' <'+maintainer[1]+'>' is a pretty-printing function or a to_string, or string_of_maintainer function.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+    parser.add_argument('--outfile', type=str, help='Output to a csv file')
+    parser.add_argument('--file', type=str, help='Select MAINTAINERS file')
+    parser.add_argument('--filter', type=str, help='Filter by file list')
+
+    args = parser.parse_args()
+
+    if args.file!= None:
+        all_maintainers_text = file_to_string(args.file)
+    else:
+        all_maintainers_text = file_to_string('./resources/linux/repo/MAINTAINERS')
```
bulwahn: is there something like a default function in python default(args.file, b) means args.file if that is not None else b.

bulwahn: Then this can be expressed less verbose.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+        counts, dirs= get_entry_counts(all_maintainers)
+        if args.filter != None:
+            counts_filt, dirs_filt = get_entry_counts(all_maintainers, filenames)
+    elif args.query == 'maintainers':
+        title = 'Maintainer'
+        counts, maintainers_subs = get_maintainer_counts(all_maintainers)
+        if args.filter != None:
+            counts_filt, _ = get_maintainer_counts(all_maintainers, filenames)
+    else:
+        return -1
```
bulwahn: Probably here you should warn about wrong use and indicate how to get some help.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+                    results.append([item, '', counts_filt[item][0], counts[item][0], counts_filt[item][1], round((counts_filt[item][0]/counts[item][0])*100, 2)])
+        elif args.filter != None and args.query == 'maintainers':
+            results.append([title, 'LoC in list', 'Total LoC', 'Byte count',  'In list(%)', 'Entries'])
+            for item in counts_filt:
+                results.append([item, counts_filt[item][0], counts[item][0], counts_filt[item][1], round((counts_filt[item][0]/counts[item][0])*100, 2), ', '.join(maintainers_subs[item]) ])
+        elif args.query == 'entries':
+            results.append([title, 'Status', 'LoC', 'Byte count'])
+            for item in counts:
+                if any(all_maintainers.subsystems[item].status):
+                    results.append([item, all_maintainers.subsystems[item].status[0].value, counts[item][0], counts[item][1]])
```
bulwahn: all_maintainers.subsystems[item].status[0].value, why is this status[0], could there actually be multiple status entries?

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,165 @@
...
+            for item in counts:
+                if any(all_maintainers.subsystems[item].status):
+                    results.append([item, all_maintainers.subsystems[item].status[0].value, counts[item][0], counts[item][1]]) 
+                else:
+                    results.append([item, '', counts[item][0], counts[item][1]])
+        else:
+            results.append([title, 'LoC', 'Byte count', 'Entries'])
+            for item in counts:
+                results.append([item, counts[item][0], counts[item][1], ', '.join(maintainers_subs[item])])
+
```
bulwahn: you can use a dictionary with lambda functions, and then rewrite this code, so that the four cases are actually just the assignment of four different lists.

bulwahn: and a generic part that handles the rest.

bulwahn (changes requested):
Here are some comments for improvements.

rralf:
@ditzlike this might be interesting for you.

bulwahn:
- how about preparing some tests to check if this command works as expected?

- for presentation of entries in show maintainers, sort entries by size themselves

and add a percentage of size of the entry compared to the overall responsibility of that maintainer.
like SCHEDULER (80%), FUTEX (15%), XYZ DRIVER (5%)

When multiple entries are there, and especially when they are cut off, this would be helpful:

(\<number of entries>) entries ...

if number of entries = 1, just drop that information.

- I did not understand "Other directories"

How about:

  - RELEVANT: list of the most general file patterns such that all files with that pattern are included in the relevant list and are in the entry.

  - IRRELEVANT: list of the most general file pattern such that all files under that directory are excluded from the relevant list, but are in the entry.

Then, I think that should cover all cases...

bulwahn:
@rralf Do you understand what we are collecting with Basak's script? Otherwise, let us quickly talk.

bin/pasta_show_maintainers.py
basakerdamar: This was to get directories, I will use a built-in function instead

bin/pasta_show_maintainers.py
basakerdamar: There seems to be always only one but returned as a list from LinuxSubsystem

rralf:
> @rralf Do you understand what we are collecting with Basak's script? Otherwise, let us quickly talk.

Hmm, I'm not sure - do you try to imitate get_maintainers.pl? BTW, @ditzlike currently writes some tests that compares output of get_maintainers with PaStA's interpretation of the MAINTAINERS file.

Is this series already ready for my review? I guess I'll understand what you're trying to achieve if I look at the code.

bulwahn:
No, we are building some basic scripts to aggregate who is "most important" based on the LoC the maintainer takes care of, and I have another script that provides the list of files for a specific kernel build, so we can create this list of most important maintainers for a kernel build.

Next step, I have in mind, is to also visualize the maintainer hierarchy based on the entries and which entry is a superset of another. Again, maybe Wolfgang already such things in the past?

bulwahn:
Ralf, you can always have a look... I will do another round of review with Basak, but we are getting closer to nice clean code in the script.

bin/pasta_show_maintainers.py
```diff
@@ -72,9 +60,11 @@ def get_maintainer_counts(all_maintainers, filenames = all_filenames):
...
             for maintainer in flatten(maintainers):
                 if not (type(maintainer) == str or maintainer[0] == '' or maintainer[1] == ''):
-                    maintainers_subs[maintainer[0]+' <'+maintainer[1]+'>'] = maintainers_subs.get(maintainer[0]+' <'+maintainer[1]+'>', set())
-                    maintainers_subs[maintainer[0]+' <'+maintainer[1]+'>'].add(subsystem)
-                    counts_by_maintainer[maintainer[0]+' <'+maintainer[1]+'>'] = np.add(counts_by_maintainer.get(maintainer[0]+' <'+maintainer[1]+'>', [0, 0]), file_lengths[filename])
+                    maintainers_subs[pretty_name(maintainer)] = maintainers_subs.get(pretty_name(maintainer), set())
+                    maintainers_subs[pretty_name(maintainer)].add(subsystem)
+                    counts_by_maintainer[pretty_name(maintainer)] = add(counts_by_maintainer.get(pretty_name(maintainer), [0, 0]), file_lengths[filename])
+                elif verbose:
+                    print('Ignoring mailing list:', maintainer)
```
bulwahn: I think pasta has a logging interface to use.

bin/pasta_show_maintainers.py
```diff
@@ -86,72 +76,68 @@ def get_small(title, counts):
...
+        percentage = lambda item : round((counts_filt[item][0]/counts[item][0])*100, 2)
+        status = lambda item : all_maintainers.subsystems[item].status[0].value if any(all_maintainers.subsystems[item].status) else ''
         
+        cases = dict()
+        cases['entries', True] =[lambda item : [item, counts_filt[item][0], counts[item][0], counts_filt[item][1], percentage(item), status(item), ', '.join(dirs[item].difference(dirs_filt[item]))], [args.query.capitalize(), 'LoC in list', 'Total LoC', 'Byte count',  'In list(%)','Status', 'Other directories']]
+        cases['entries', False] =[lambda item : [item, counts[item][0], counts[item][1], status(item)], [args.query.capitalize(), 'LoC', 'Byte count','Status']]
+        cases['maintainers', True] =[lambda item : [item, counts_filt[item][0], counts[item][0], counts_filt[item][1], percentage(item), ', '.join(maintainers_subs[item])], [args.query.capitalize(), 'LoC in list', 'Total LoC', 'Byte count',  'In list(%)', 'Entries']]
+        cases['maintainers', False] =[lambda item : [item, counts[item][0], counts[item][1]], [args.query.capitalize(), 'LoC', 'Byte count']]
+
```
bulwahn: Now, look what are duplicate lambda functions and what is the configurable part here? If you split it accordingly, you got a much nicer solution.

bin/pasta_show_maintainers.py
```diff
@@ -86,72 +76,68 @@ def get_small(title, counts):
...
+        if 'counts_filt' not in locals():
```
bulwahn: I think that is a bit too much of meta-programming here. Consider just having one variable above.

bin/pasta_show_maintainers.py
bulwahn: I actually meant a dictionary for 'field_name' => function, so 'LoC' => lambda item: counts[item][0]... etc.

bulwahn (changes requested):
We are getting there...

rralf:
Let's Cc Wolfgang. @wolfgangmauerer 

I understand. Two things before my first round of review:
  - If we only need 'add' from numpy, then I'd like to substitute it with something else. Numpy is a beast of a dependency that we probably can avoid, as long as there's no strong reason for it.
  - Basak, could you please rebase on next, and reorder your commits? E.g., first add the stubs to pasta, then add the actual business logic. Makes it easier for me to review

Thanks!
  Ralf

bin/pasta_show_maintainers.py
basakerdamar: >     * I did not understand "Other directories"
> 
> 
> How about:
> 
>     * RELEVANT: list of the most general file patterns such that all files with that pattern are included in the relevant list and are in the entry.
> 
>     * IRRELEVANT: list of the most general file pattern such that all files under that directory are excluded from the relevant list, but are in the entry.
> 
> 
> Then, I think that should cover all cases...

Hello, 
Other directories are showing the directories that are excluded in the filter list for each of the entries. Would it be meaningful to display some aggregation of these directories in this case?

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,174 @@
+"""
+PaStA - Patch Stack Analysis
+Copyright (c) Bayerische Motoren Werke Aktiengesellschaft (BMW AG), 2019
```
bulwahn: It is 2020 already, either change it to 2020 or just drop it :)

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,174 @@
...
+    fields['Maintainers'] = fields['Entries'] = lambda item : item
+    fields['LoC in list'] = lambda item : counts[item][2]
+    fields['Total LoC'] = fields['LoC'] = lambda item : counts[item][0]
+    fields['Byte count'] = lambda item : counts[item][1]
+    fields['In list(%)'] = lambda item : round((counts[item][2]/counts[item][0])*100, 2)
+    fields['Status'] = lambda item : all_maintainers.subsystems[item].status[0].value if any(all_maintainers.subsystems[item].status) else ''
+    fields['Other directories'] = lambda item : ', '.join(dirs[item])
+    fields['Entries of maintainer'] = lambda item : ', '.join(maintainers_subs[item])
+       
+    if args.smallstat:
```
bulwahn: If you look how these lists are composed, you will find that you can compose this list one by one nicely.
E.g.,
title = [a, b if condition, c, d if condition, e if condition, ...]

bulwahn (commented):
Some more feedback

basakerdamar:
> Could you please provide an example how I can exemplarily run this script?

Hello, 

 ./pasta show_maintainers --entries

should show you a view with subsystems and respective sums of file sizes and

./pasta show_maintainers

brings a similar view, this time with maintainers.

You could give a file containing a list with --filter so that the ratios in-list/in-repo are also displayed. In the filter file, file paths should be relative to the repository like arch/x86/boot/a20.c.

I changed the commands a bit for the sake of simplicity. I hope the script also looks a bit better with the improvements.

rralf:
Great, thank you very much! Will try that tomorrow.

rralf:
> > Could you please provide an example how I can exemplarily run this script?
> 
> Hello,
> 
> ./pasta show_maintainers --entries
> 
> should show you a view with subsystems and respective sums of file sizes and
> 
> ./pasta show_maintainers
> 
> brings a similar view, this time with maintainers.
> 
> You could give a file containing a list with --filter so that the ratios in-list/in-repo are also displayed. In the filter file, file paths should be relative to the repository like arch/x86/boot/a20.c.
> 
> I changed the commands a bit for the sake of simplicity. I hope the script also looks a bit better with the improvements.

Basak, thanks for the pointer, I think begin to understand what you try to achieve.

bulwahn:
Basak, the feature listing 

RELEVANT: list of the most general file patterns such that all files with that pattern are included in the relevant list and are in the entry.

IRRELEVANT: list of the most general file pattern such that all files under that directory are excluded from the relevant list, but are in the entry.

is still missing, right?

If it is not clear at the moment, let us just the dump the whole list of files included and excluded and go through a few examples to see how that can be nicely aggregated.

basakerdamar:
For both RELEVANT and IRRELEVANT, I was implementing most common containing folder(s) but it did not seem like a smart solution as an aggregation. I will bring back the list of directories for now.

bulwahn:
Okay, I will have a look.

bulwahn:
Can we include the list of files that are considered relevant and irrelevant?

I think there is some kind of bug in showing the directories; I simply could not make sense of why it was either in relevant or irrelevant when it was actually mixed when cross-checking the list of relevant files and the description in MAINTAINERS.

bulwahn:
I think we can also drop the byte count by default, it confuses me more than it helps.

rralf (commented):
The files that are affected by a specific configuration should go to the resources/ directory. @bulwahn It'd also be great if we could include the script that creates the file.

Ah, and we could simply us the file extension .files, and think of a standardises naming scheme. We can use information of the filename to tell PaStA which version of MAINTAINERS it should load. Basak, your filename already contains "v5.6", which is very useful.

We could (just a suggestion) use a dot-separation like:

config-whatever-x86-machine.v5.5-rc3.files

For reasons traceability, I'd also like to store the config itself.

rralf:
For the rest of the PR: Could you please make a clean patch stack out of it? Makes it easier for me to review. It'd be great if there would be just two commits: The first one that introduces the stubs for the new tool, and a second one that adds the business logic on top. The second one should introduce a working state of your tool. Any further changes/improvements can be made on top of that.

pasta
```diff
@@ -43,7 +43,7 @@ from bin.pasta_patch_descriptions import patch_descriptions
 from bin.pasta_upstream_duration import upstream_duration
 from bin.pasta_upstream_history import pasta_upstream_history
 from bin.pasta_web import web
-from bin.pasta_show_maintainers import get_maintainers
+from bin.pasta_show_maintainers import show_maintainers
```
rralf: No need to change this line, if you use the final name (show_maintainers) in the first commit.

pasta
```diff
@@ -171,8 +171,8 @@ def main(argv):
         return upstream_duration(config, sub, argv)
     elif sub == 'web':
         return web(config, sub, argv)
-    elif sub == 'maintainers':
-        return get_maintainers(config, sub, argv)
+    elif sub == 'show_maintainers':
+        return show_maintainers(config, sub, argv)
```
rralf: same here

rralf (changes requested):

rralf (commented):
Please use the final name show_maintainers in this commit

rralf:
Thanks! Will have a deeper look into the second commit tomorrow!

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+    fields['Irrelevant directories'] = lambda item: ', '.join(irrelevant_dirs[item[0]])
+    fields['Relevant directories'] = lambda item: ', '.join(relevant_dirs[item[0]])
+   
+    title.append('LoC')
+    if not args.smallstat:
+        if args.filter != None:
+            title += ['LoC in list', 'In list(%)']
+            if args.entries:
+                title += ['Irrelevant directories', 'Relevant directories']
+        if args.bytes:
```
bulwahn: If you move this a bit above, you can simply use append or title += ['Byte count'].

bulwahn:
> The files that are affected by a specific configuration should go to the resources/ directory. @bulwahn It'd also be great if we could include the script that creates the file.
>

On my todo list for weeks now, it is a simple script, but I need to document it properly.
 
> Ah, and we could simply us the file extension .files, and think of a standardises naming scheme. We can use information of the filename to tell PaStA which version of MAINTAINERS it should load. Basak, your filename already contains "v5.6", which is very useful.
> 
> We could (just a suggestion) use a dot-separation like:
> 
> config-whatever-x86-machine.v5.5-rc3.files
> 
> For reasons traceability, I'd also like to store the config itself.

We can surely do that.

bulwahn:
> Any further changes/improvements can be made on top of that.

I would still like to see the relevant and irrelevant files/directories as a feature in a way it is useful for me. We can commit it to the repository in the current state, but I still think it needs some improvements (or different options to be able to query what the user really wants to know) afterwards.

rralf (changes requested):
You just deleted the resources submodule. What i meant was to move the config to the resources submodule.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+    parser.add_argument('--file', type=str, help='Specify MAINTAINERS file')
+    parser.add_argument('--filter', type=str, help='Filter by file list')
+    parser.add_argument('--commit', type=str, 
+                            help='Specify a commit hash for Linux repo')
+    parser.add_argument('--cpu', dest='cpu_factor', metavar='cpu', type=float,
+                        default=1.0, help='CPU factor for parallelisation '
+                                          '(default: %(default)s)')
+    args = parser.parse_args()
+    d_repo = config.repo_location
+    repo = pygit2.Repository(d_repo)
```
rralf: repo = config.repo.repo (we already have a Repository instance)

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+        filter_by_files = file_to_string(args.filter).splitlines()
+        # Automatically find kernel version from file name "v5.6"
+        version_regex = re.compile(r'v[0-9]+.[0-9]+')
+        kernel_version = version_regex.findall(args.filter)[0]
+
+    if args.file:
+        all_maintainers = file_to_string(args.file)
+        LinuxMaintainers(all_maintainers)
+    else:
+        if kernel_version:
```
rralf: if !args.filter -> referenced before assignment

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+                            help='Specify a commit hash for Linux repo')
+    parser.add_argument('--cpu', dest='cpu_factor', metavar='cpu', type=float,
+                        default=1.0, help='CPU factor for parallelisation '
+                                          '(default: %(default)s)')
+    args = parser.parse_args()
+    d_repo = config.repo_location
+    repo = pygit2.Repository(d_repo)
+    commit_version = args.commit if args.commit else 'HEAD'
+    commit = repo.revparse_single(commit_version)
+    tree = commit.tree
```
rralf: Hmm... Fails if I use --commit v5.3. Would be nice if we could support both.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+                            help='groups by subsystems')
+    parser.add_argument('--verbose', action='store_true', 
+                            help='Show ignored exceptions on the way')
+    parser.add_argument('--smallstat', action='store_true', help='Simple view')
+    parser.add_argument('--bytes', action='store_true', help='Show byte counts')
+    parser.add_argument('--outfile', type=str, help='Output to a csv file')
+    parser.add_argument('--file', type=str, help='Specify MAINTAINERS file')
+    parser.add_argument('--filter', type=str, help='Filter by file list')
+    parser.add_argument('--commit', type=str, 
+                            help='Specify a commit hash for Linux repo')
```
rralf: Change --commit to --revision. --revision should at least support a specific commit hash and a tree object. And we should properly handle errors.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+                                            'maintainers or subsystems')
+    parser.add_argument('show_maintainers', nargs=1, type=bool)
+    parser.add_argument('--entries', action='store_true', 
+                            help='groups by subsystems')
+    parser.add_argument('--verbose', action='store_true', 
+                            help='Show ignored exceptions on the way')
+    parser.add_argument('--smallstat', action='store_true', help='Simple view')
+    parser.add_argument('--bytes', action='store_true', help='Show byte counts')
+    parser.add_argument('--outfile', type=str, help='Output to a csv file')
+    parser.add_argument('--file', type=str, help='Specify MAINTAINERS file')
```
rralf: Do we still need that argument? I guess we can drop it. It's totally sufficient if we can specify the kernel version.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+    parser.add_argument('--filter', type=str, help='Filter by file list')
+    parser.add_argument('--commit', type=str, 
+                            help='Specify a commit hash for Linux repo')
+    parser.add_argument('--cpu', dest='cpu_factor', metavar='cpu', type=float,
+                        default=1.0, help='CPU factor for parallelisation '
+                                          '(default: %(default)s)')
+    args = parser.parse_args()
+    d_repo = config.repo_location
+    repo = pygit2.Repository(d_repo)
+    commit_version = args.commit if args.commit else 'HEAD'
```
rralf: If we encode the version in the filename, then we should use this information as the default value instead of HEAD. --revision can be used to override the version.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+                config, [kernel_version]).get(kernel_version)
+        else:
+            all_maintainers = tree['MAINTAINERS'].data.decode('iso8859')
+
+    all_filenames = walk_commit_tree(tree)
+
+    global _tmp_tree
+    _tmp_tree = tree
+
+    processes = int(cpu_count() * args.cpu_factor)
```
rralf: Did you experience any scaling issues? If not, then I'd say we can drop the cpu_factor.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+        columns = int(columns)
+        for line in results:
+            line_s = str(line[0]).ljust(width+4)
+            for item in line[1:]:
+                line_s += str(item).ljust(12)
+            if len(line_s) > columns:
+                print(line_s[0:columns-3]+"...")
+            else:
+                print(line_s)
+    return 0
```
rralf: Nit: Missing newline at EOF. And please keep an eye on trailing spaces. You can, for example, use git show to detect them.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+        
+    if args.filter:
+        filter_by_files = file_to_string(args.filter).splitlines()
+        # Automatically find kernel version from file name "v5.6"
+        version_regex = re.compile(r'v[0-9]+.[0-9]+')
+        kernel_version = version_regex.findall(args.filter)[0]
+
+    if args.file:
+        all_maintainers = file_to_string(args.file)
+        LinuxMaintainers(all_maintainers)
```
rralf: ... Looks incomplete. However, I guess we can drop it entirely.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+    parser.add_argument('show_maintainers', nargs=1, type=bool)
+    parser.add_argument('--entries', action='store_true', 
+                            help='groups by subsystems')
+    parser.add_argument('--verbose', action='store_true', 
+                            help='Show ignored exceptions on the way')
+    parser.add_argument('--smallstat', action='store_true', help='Simple view')
+    parser.add_argument('--bytes', action='store_true', help='Show byte counts')
+    parser.add_argument('--outfile', type=str, help='Output to a csv file')
+    parser.add_argument('--file', type=str, help='Specify MAINTAINERS file')
+    parser.add_argument('--filter', type=str, help='Filter by file list')
```
rralf: Please try to improve the description of this argument. I had to walk through the code in order to understand, that --filter needs to point to the defconfig.files file.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+    d_repo = config.repo_location
+    repo = pygit2.Repository(d_repo)
+    commit_version = args.commit if args.commit else 'HEAD'
+    commit = repo.revparse_single(commit_version)
+    tree = commit.tree
+        
+    if args.filter:
+        filter_by_files = file_to_string(args.filter).splitlines()
+        # Automatically find kernel version from file name "v5.6"
+        version_regex = re.compile(r'v[0-9]+.[0-9]+')
```
rralf: Simply split the filename at the dots. Then, there's no need for a regex (that does, for example, currently not work for v5.4-rc3).

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,299 @@
...
+        all_maintainers = file_to_string(args.file)
+        LinuxMaintainers(all_maintainers)
+    else:
+        if kernel_version:
+            all_maintainers = load_maintainers(
+                config, [kernel_version]).get(kernel_version)
+        else:
+            all_maintainers = tree['MAINTAINERS'].data.decode('iso8859')
+
+    all_filenames = walk_commit_tree(tree)
```
rralf: Ok, just a second. Let's get one step back and think of our initial problem:

We want to know the scale/volume of subsystems of the kernel. Either from an overall perspective, or related to a specific configuration. A specific configuration is always mapped to a version and list of affected files. So it is not necessary to specify both: Kernel version and a commit hash for the commit tree walk.

Let me give you two examples:

Get an overall overview of the maintainers of v5.3
$ ./pasta show_maintainers --entries --revision v5.3

Get an overall overview of the maintainers of HEAD
$ ./pasta show_maintainers --entries

Get an overview of the maintainers for a specific configuration, will default to MAINTAINERS of v5.4
$ ./pasta show_maintainers --entries --config defconfig-x86.v5.4.files

Get an overview of the maintainers for a specific configuration, will enforce v5.5, though (probably never required)
$ ./pasta show_maintainers --entries --config defconfig-x86.v5.4.files --revision v5.5

rralf (changes requested):
Hi Basak,

thank you very much for your work! I can see the improvements in every round. So please don't be confused that there are still so many comments. That's not unusual. In this round, I only focused on the configuration. I did not dive into other parts of your script. I will do that once we have a common understanding of how to run/configure the script.

Thanks
  Ralf

bin/pasta_show_maintainers.py
```diff
@@ -193,42 +204,35 @@ def show_maintainers(config, sub, argv):
...
-    commit = repo.revparse_single(commit_version)
-    tree = commit.tree
-        
+
     if args.filter:
         filter_by_files = file_to_string(args.filter).splitlines()
         # Automatically find kernel version from file name "v5.6"
-        version_regex = re.compile(r'v[0-9]+.[0-9]+')
-        kernel_version = version_regex.findall(args.filter)[0]
+        extract_version = '.'.join(args.filter.split('.')[1:3])
```
rralf: Yikes, I just realised that dot is probably not the best choice as separator :-)

rralf: Let me try another approach, will get back to you.

basakerdamar: That was just my work in progress, sorry about the mess :) 
Should I go back to regex?

rralf: No worries. I'm not sure let. I'm playing around with some possibilities. Will provide you a patch for on top of your series.

basakerdamar:
I think that I used the wrong command to move the relevant-files which ended up deleting the resources submodule. Sorry about the confusion! I will fix that now.

rralf:
Hold on, I've already a local workaround. Will push it soon.

rralf:
So I just implemented 8a4cd618450c4282cc38893c6dc574e298b5929b and 4ec2d6dd82c585eb1795b72ea3f0828752d0d1b8. You might want to have a look at them. Give me a couple of minutes, I'll rebase your latest push and add my local changes.

rralf:
Ok, so I made some API changes that are already integrated on the next branch. Those API changes make it much easier to load a LinuxMaintainers instance given by any revision. You could do, for example:

mtrs = load_maintainers(repo, 'v5.4-rc3')
mtrs = load_maintainers(repo, '1a2b3c4d5d6e')
mtrs = load_maintainers(repo, 'master')
mtrs = load_maintainers(repo, 'HEAD')

They all will work. This makes life a lot easier. I used this chance to implement a get_blob routine in the Repository class. It takes a revision and a filename and returns a blob. Can be useful for many cases.

Given that, have a look at this branch:
https://github.com/lfd/PaStA/tree/basak

It's your tree, rebased to latest next. I did following changes:

- The version is no longer encoded in the filename. It was a silly idea (by me).
- The version is now encoded in the file itself: The version is simply the first line
- --revision can always override the version, even if --filter is given
- If --revision is not set, the script takes the first line of the file
- If --filter is not set, it will default to HEAD
- It will finally log which version it chose

See 0320d549d8f07da8aac278c561aea1b897546f28.

basakerdamar:
> Basak, the feature listing
> 
> RELEVANT: list of the most general file patterns such that all files with that pattern are included in the relevant list and are in the entry.
> 
> IRRELEVANT: list of the most general file pattern such that all files under that directory are excluded from the relevant list, but are in the entry.
> 
> is still missing, right?
> 
> If it is not clear at the moment, let us just the dump the whole list of files included and excluded and go through a few examples to see how that can be nicely aggregated.

I added my changes regarding this feature. I removed my old aggregation which lists the common (sub-)directories instead of listing everything, as it was causing confusion when cross-checking relevant and irrelevant lists. Right now, the script plainly lists directories in the filter file list and out of the filter file list.

rralf:
Hi Basak,

thanks for the changes! Could you please rebase your branch to the latest version of the next branch of PaStA? We made a lot of changes upstream in the meanwhile.

Thanks!

rralf:
Hi Basak,

that was a merge with next. The history is now pretty confusing. Please [rebase](https://git-scm.com/docs/git-rebase) to the latest next branch, so that we get a linear history, iow, w/o criss-cross merges. You can also squash / remove commits that we don't need any longer.

Thanks!

basakerdamar:
Hi Ralf, 
Thank you for your help. I re-ordered and squashed my commits so that there is one commit for stubs to pasta and one commit with the script itself, like it was before.

rralf:
Perfect, thanks. Makes it much easier for me to review. Will have a look at it tomorrow.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+Copyright (c) OTH Regensburg, 2019-2020
+Authors:
+  Basak Erdamar <basakerdamar@gmail.com>
+This work is licensed under the terms of the GNU GPL, version 2. See
+the COPYING file in the top-level directory.
+"""
+
+import sys
+import os
+import pygit2
```
rralf: nit: please order alphabetically

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+
+from csv import writer
+from logging import getLogger
+from argparse import ArgumentParser
+from multiprocessing import Pool, cpu_count
```
rralf: same here

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+log = getLogger(__name__[-15:])
+
+# We need this global variable, as pygit2 Commit.tree objects are not pickleable
+_tmp_tree = None
+
+def flatten(nested_list):
+    return [item for sublist in nested_list for item in sublist]
+
+def walk_commit_tree(tree):
+    results = []
```
rralf: An object can't appear twice in the list. So
results = set()
would be the appropriate container.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+    for entry in tree:
+        if type(entry) == pygit2.Blob:
+            results.append(entry.name)
+        elif type(entry) == pygit2.Tree:
+            results += [os.path.join(entry.name, item)
+                        for item in walk_commit_tree(entry)]
+    return results
+
+def get_tree(repo, revision):
+    if revision != 'HEAD':
```
rralf: Please reverse the logic:

```
if revision == 'HEAD':
    return repo.revparse_single(repo.head.name).tree

if revision.startswith...
```

Makes it easier to read.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+        try:
+            if revision.startswith('v'):
+                revision = repo.lookup_reference('refs/tags/%s' % revision).target
+            commit_hash = repo[revision].target
+            tree = repo[commit_hash].tree
+        except ValueError:
+            raise
```
rralf: If we just raise the exception again, the the try...except block becomes superfluous. Remove it.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+    else:
+        commit = repo.revparse_single(repo.head.name)
+        tree = commit.tree
+    return tree
+
+def decreasing_order(dictionary):
+    return sorted(dictionary.items(), key=lambda item:item[1].loc, reverse=True)
+
+def read_file(filename):
+    return _tmp_tree[filename].data.decode('iso8859')
```
rralf: single-caller function. please fold it into get_filesize as long as there are no other users.

rralf: And why do you use iso8859? What if a file is encoded in UTF-8? There are quite some.

But if you just need to count for line feeds, there's no need to decode the bytesarray at all.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+
+def get_filesize(filename):
+    f = read_file(filename)
+    try:
+        return (filename, (f.count('\n'), len(f)))
```
rralf: instead of len(f), you should use _tmp_tree[data].size

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+    # Empty file:
+    except:
+        return (filename, (0, 0))
```
rralf: When does an exception occur? This looks odd to me.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+def get_counts_by_maintainer(all_maintainers, file_sizes,
+                                verbose, filter_by_files = None):
+    maintainer_to_subsystem = dict()
+    # maps maintainers to total file sizes they are responsible for
+    counts = dict()
+    for filename in file_sizes:
+        subsystems = all_maintainers.get_subsystems_by_file(filename)
+        for subsystem in subsystems:
+            maintainers = all_maintainers.get_maintainers(subsystem)
+            for maintainer in flatten(maintainers):
```
rralf: Uaah, no! :-)

This is not correct.

all_maintainers.get_maintainers(subsystem)

[get_maintainers](https://github.com/lfd/PaStA/blob/master/pypasta/LinuxMaintainers.py#L157) _always_ gives you a tripe (lists, maintainers, reviewers). I'm still not quite sure what you're trying to achieve, but probably, you only want to access the maintainers.

(I have to admin, maybe get_maintainers should be refactored to get_responsible or something like that)

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+                if not (type(maintainer) == str
+                        or maintainer[0] == ''
+                        or maintainer[1] == ''):
```
rralf: Then, this become superfluous.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+    else:
+        width = len(max(list(zip(*results))[0], key = len))
+        _, columns = os.popen('stty size', 'r').read().split()
+        columns = int(columns)
+        for line in results:
+            line_s = str(line[0]).ljust(width+4)
+            for item in line[1:]:
+                line_s += str(item).ljust(12)
+            if len(line_s) > columns:
+                print(line_s[0:columns-3]+"...")
```
rralf: Ok, I don't know what kind of magic happens here, but we first need to address the points above.

bin/pasta_show_maintainers.py
```diff
@@ -0,0 +1,286 @@
...
+            self.loc_filt = self.loc
+            self.byte_filt = self.byte
+        else:
+            self.loc_filt = self.byte_filt = 0
+    def increase(self, filename, loc, byte, filter):
+        self.loc += loc
+        self.byte += byte
+        if filter:
+            self.loc_filt += loc
+            self.byte_filt += byte
```
rralf: Ok, lots of magic that I don't understand yet, but we need to solve the other issues first.

rralf (changes requested):
Hi Basak,

thanks for the new version!

Please find my review attached. I already addressed some of the comments in order to get the script running. Please find those patches on the [basak](https://github.com/lfd/PaStA/tree/basak)-branch. You can integrate the topmost patch to your branch.

Most important thing: I guess you misunderstood the result of get_maintainers(). It returns you a triple of lists, maintainers, reviewers. You try to somehow (erroneously) flatten(?) the result.

Besides that, I tried to run `pasta show_maintainers --filter test` with the following content of test:
```
v4.7-rc3
MAINTAINERS
scripts/get_maintainers.pl
Makefile
```

As result, I got:
```2020-06-02 22:54:51,096 how_maintainers INFO     Working on kernel revision v4.7-rc3
Maintainers                                        LoC         LoC in list In list(%)  Entries of maintainer
linus torvalds  <torvalds@linux-foundation.org>    21716623    14399       0.07        THE REST    
michal marek  <mmarek@suse.com>                    13786       1667        12.09       COCCINELLE/Semantic Patches (SmPL), KERNEL BUILD + files below scripts...
2020-06-02 22:56:43,880 PaStA           INFO     Shutting down


```

... which somehow looks odd to me. Is the output intended?


One last thing: I'm somehow struggling with the name `show_maintainers`. I bet we can find a better name for the script. What we actually do is printing statistics of responsibilities of either the whole kernel, or a certain subset of the kernel. Maybe a better name would be `subsystem_stats` or maybe `maintainers_stats`. I happily accept better suggestions! :-)

Thanks!
  Ralf

rralf:
https://github.com/lfd/PaStA/commit/cee96b78127eb950e09768107f890c3b85b3670c

So this is the commit with my fixups. Mainly, I improved walk_commit_tree (use sets instead of lists), simplified get_tree() and corrected get_file_size() (we don't need decoding of strings, and we mustn't use len() to get the size of a file).

bin/pasta_show_maintainers.py
basakerdamar: I meant this to be a counter for lines, keeping track of in-list and and total separately. Though I do not need the file name anymore, I fixed this in my last push.

basakerdamar:
> Hi Basak,
> 
> thanks for the new version!
> 
> Please find my review attached. I already addressed some of the comments in order to get the script running. Please find those patches on the [basak](https://github.com/lfd/PaStA/tree/basak)-branch. You can integrate the topmost patch to your branch.
> 
> Most important thing: I guess you misunderstood the result of get_maintainers(). It returns you a triple of lists, maintainers, reviewers. You try to somehow (erroneously) flatten(?) the result.
> 
> Besides that, I tried to run `pasta show_maintainers --filter test` with the following content of test:
> 
> ```
> v4.7-rc3
> MAINTAINERS
> scripts/get_maintainers.pl
> Makefile
> ```
> 
> As result, I got:
> 
> ```
> Maintainers                                        LoC         LoC in list In list(%)  Entries of maintainer
> linus torvalds  <torvalds@linux-foundation.org>    21716623    14399       0.07        THE REST    
> michal marek  <mmarek@suse.com>                    13786       1667        12.09       COCCINELLE/Semantic Patches (SmPL), KERNEL BUILD + files below scripts...
> 2020-06-02 22:56:43,880 PaStA           INFO     Shutting down
> ```
> 
> ... which somehow looks odd to me. Is the output intended?
> 
> One last thing: I'm somehow struggling with the name `show_maintainers`. I bet we can find a better name for the script. What we actually do is printing statistics of responsibilities of either the whole kernel, or a certain subset of the kernel. Maybe a better name would be `subsystem_stats` or maybe `maintainers_stats`. I happily accept better suggestions! :-)
> 
> Thanks!
> Ralf

Hi Ralf,

Thank you for the changes!

The output was what I intended, although now 'THE REST' is excluded in each view. Hopefully this way it will be more meaningful.

I also added the --files option to see the entries for each file and changed the script's name to maintainers_stats.

rralf:
Then you need to explain the output to me. There are surely more maintainers than just Michal Marek that are responsible for the three files.

rralf:
Okay, I still don't get it. What's the semantics of `maintainers_state --filter`? What does it tell me? With your new version, it prints now nothing (same example as above):
```
2020-06-04 12:48:44,590 PaStA           INFO     Cmdline: ./pasta maintainers_stats --filter ./test
2020-06-04 12:48:44,590 pypasta.Config  INFO     Active configuration: linux
2020-06-04 12:48:44,635 intainers_stats INFO     Working on kernel revision v4.7-rc3
Maintainers    LoC         LoC in list In list(%)  Entries of maintainer
2020-06-04 12:51:02,258 PaStA           INFO     Shutting down
```

basakerdamar:
> Then you need to explain the output to me. There are surely more maintainers than just Michal Marek that are responsible for the three files.

Sorry, I thought the confusion was about the format/fields of the output. There are clearly some lines missing in this result. I am trying to find the problem causing the missing maintainers, I pushed another fix too.

bin/pasta_maintainers_stats.py
```diff
@@ -0,0 +1,303 @@
...
+    file_sizes = dict(result)
+
+    if args.entries:
+        title = ['Entries','LoC']
+        if args.filter:
+            results, irrelevant_dirs, relevant_dirs = get_counts_by_entry(
+                all_maintainers, file_sizes, filter_by_files)
+        else:
+            results, _, _= get_counts_by_entry(all_maintainers, file_sizes)
+    elif args.files:
```
rralf: `--entries` and `--files` are mutual exclusive, right?

So if I call `maintainers-stats --entries --files`, then --files will by ignored.

In this case, it's probably better to make it clear by introducing a mode. E.g.
```
maintainers_stats --mode entries
maintainers_stats --mode files
```

We even have three modes, right? entries, files, and maintainers, if i see that correctly. We should make that clear in the commandline arguments.

And it's probably better to rename entries to subsystems.

bin/pasta_maintainers_stats.py
```diff
@@ -0,0 +1,303 @@
...
+        # If filter is given, only return the relevant lines:
+        counts = {name:item for name,item in counts.items()
+                    if item.display_in_filter()}
+    # Sort maintainers by decreasing total line counts
+    return decreasing_order(counts), maintainer_to_subsystem
+
+
+def get_entries_by_files(all_maintainers, filenames):
+    filenames = sorted(filenames, key=lambda item:os.path.basename(item), 
+                        reverse=True)
```
rralf: Could you explain what this routine does? I don't unterstand the option --files.

rralf (changes requested):
Hi Basak,

cool, things are getting much clearer now! Again, have a look at the upstream [basak branch](https://github.com/lfd/PaStA/tree/basak). I added two commits that simplify things.

The first commit simplifies the logic inside the evaluators (we can drop a lot of duplicated code to simplify things), and the second commit simplifies the structure of the caller-site.

However, we aren't yet there. What I realised during the refactoring is that get_counts_by_entry and get_counts_by_maintainer basically almost have the same structure. We should consolidate those routines to one single routine, and add an additional argument that selects wether the user wants to consider maintainer or entries.

BTW: With entries, you refer to entries in MAINTAINERS, aka. subsystems, right? If so, we should use the term subsystem instead of entry.

I also made some changes to the Counter-class. Please check if I missed something, I'm still not totally aware of its internal logic.

Thanks!
  Ralf

bulwahn:
@basakerdamar, here is some feedback to --files option on fdcbc61:
  - order files alphabetically by default
  - how are the entries ordered? Can we order them by size by default?
  - let us drop LoC, file size by default and create an option for adding those columns if requested.

bulwahn:
> BTW: With entries, you refer to entries in MAINTAINERS, aka. subsystems, right? If so, we should use the term subsystem instead of entry.
> 
Yes, with entries we refer to entries in MAINTAINERS and I disagree with the suggestion to use the term subsystem. Not every entry in MAINTAINERS is a subsystem (in kernel developer lingo). Look at the wording in MAINTAINERS, we can use "section" (as in "Descriptions of section entries" in MAINTAINERS, https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/MAINTAINERS#n80) or "entries" as in  "When adding to this list, please keep the entries in alphabetical order.", https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/MAINTAINERS#n142). Hence, I early decided for entries in my wording and kernel developers did not disagree on my patches on MAINTAINERS. So, either section or entry, as far as I see it. Entry is not to bad for this tool.

bin/pasta_maintainers_stats.py
bulwahn: Very quickly, it gives you a list of file to entry mapping.

Here is quick dump of the csv output, dumped into Google Sheets and formated and reordered (alphabetically): https://docs.google.com/spreadsheets/d/13O5bstZpENC04EvrTziYKe-DilyLwTH6RQ7sWukS82c/edit?usp=sharing

You can quickly see there is structure and substructure of responsibilities according to the directory etc., maybe we then also find way to visualize that nicely...

rralf:
Ok, but then `get_maintainers.pl --subsystem` should then consistently be changed to `get_maintainers.pl --entries` instead.

I get your point, but from my understanding, and in terms of get_maintainers.pl, the main "user" of MAINTAINERS, a section in MAINTAINERS is a subsystem. For me, the term 'entry' is a bit misleading, as I rather associate a single line with it instead of a whole section.

In the document that you pointed to, they mention the term 'section entry'. I'd rather use this term, or the short variant `section` then.

bin/pasta_maintainers_stats.py
bulwahn: Yes, we should introduce some option like that, I would call it --group-by or --aggregate-on ... entries (or section, not subsystem)/files/maintainers instead of --mode.

bulwahn:
> Ok, but then `get_maintainers.pl --subsystem` should then consistently be changed to `get_maintainers.pl --entries` instead.
>
Consistency, yes!
 
> I get your point, but from my understanding, and in terms of get_maintainers.pl, the main "user" of MAINTAINERS, a section in MAINTAINERS is a subsystem. For me, the term 'entry' is a bit misleading, as I rather associate a single line with it instead of a whole section.
>

But simply go out and say you work on the CLANG-FORMAT FILE subsystem, people are going to say that is not a subsystem, although it deserve to be a section in MAINTAINERS.
  
> In the document that you pointed to, they mention the term 'section entry'. I'd rather use this term, or the short variant `section` then.

Okay, I am happy to settle for section. Let us get the whole wording right and consistent for the commandline and the whole code with using the word section.

bin/pasta_maintainers_stats.py
rralf: Thanks, got it. Then I'd like to ask for the following change:

AFAICT, the ordering of the filenames does not matter for the evaluation, right? You alphabetically order filenames for reasons of visualisation.

Could you please remove the ordering from this routine? Ordering is for visualisation and not for processing. So we should, for example, return a dictionary {filename: ([entries], loc, size)}. The caller can then do further postprocessing, like ordering things.
The reason: I was wondering why the order plays a role while iterating over the filenames. If we postpone the ordering, it gets clear that it's just about visualisation.

bin/pasta_maintainers_stats.py
rralf: Yep, sounds good. And use, e.g., sections (formerly knows as entries) as default value.

basakerdamar:
> Then you need to explain the output to me. There are surely more maintainers than just Michal Marek that are responsible for the three files.

Hi Ralf,
Short question for debugging:
For your filter example,
from 
```
LinuxMaintainers.get_subsystems_by_file('Makefile')
```
I get:
```
{'THE REST', 'KERNEL BUILD + files below scripts/ (unless maintained elsewhere)'}
```
and
```
LinuxMaintainers.get_maintainers('KERNEL BUILD + files below scripts/ (unless maintained elsewhere)')
```
gives
```
({'linux-kbuild@vger.kernel.org'}, [('michal marek ', 'mmarek@suse.com')], [])
```
Same for scripts/get_maintainer.pl, I only get one maintainer:
```
{'GET_MAINTAINER SCRIPT', 'THE REST'}
```
and for 'GET_MAINTAINER SCRIPT', 
```
(set(), [('joe perches ', 'joe@perches.com')], [])
```
The results is therefore only two maintainers. Could it be that LinuxMaintainers class gives me too few maintainers or am I using the class in a wrong way?

bin/pasta_maintainers_stats.py
```diff
@@ -0,0 +1,275 @@
...
+                                            'maintainers or sections')
+    parser.add_argument('maintainers_stats', nargs=1, type=bool)
+    parser.add_argument('--smallstat', action='store_true', help='Simple view')
+    parser.add_argument('--bytes', action='store_true', help='Show byte counts')
+    parser.add_argument('--outfile', type=str, help='Output to a csv file')
+    parser.add_argument('--filter', type=str, help='Filter by file list: '
+                        'enter the file name for the file containing the list '
+                        'of files to filter by.')
+    parser.add_argument('--group-by', type=str, default='sections',
+                        help='files option '
```
rralf, line 115: Add the choices={'entries', 'section', 'files'} argument. argparse will do then sanity checks for arguments.

bin/pasta_maintainers_stats.py
```diff
@@ -0,0 +1,275 @@
...
+        title = ['Maintainers', 'LoC']
+        results, maintainer_to_section = \
+         get_counts(all_maintainers, file_sizes, filter_by_files, args.group_by)
+    elif args.group_by == 'files':
+        title = ['File name', 'Sections of file']
+        victims = filter_by_files if len(filter_by_files) else all_filenames
+        results = get_sections_by_files(all_maintainers, victims)
+    else:
+        parser.print_help()
+        sys.exit()
```
rralf: We can leave out the whole else case if we feed the argparser with the choices argument

bin/pasta_maintainers_stats.py
```diff
@@ -0,0 +1,275 @@
...
+                        'of files to filter by.')
+    parser.add_argument('--group-by', type=str, default='sections',
+                        help='files option '
+                        'Displays files with corresponding sections\n'
+                        'sections option groups files by sections, displays '
+                        'sections ordered by lines of code they include\n'
+                        'maintainers option groups files by sections, displays '
+                        'each maintainer ordered by lines of code they are '
+                        'responsible for\n'
+                        'Default is sections if no value is specified')
```
rralf, line 122: Please separate sentences with interpunctation. And have a look at the output that pasta maintainers_stats -h generates, this help section is really hard to read.

bin/pasta_maintainers_stats.py
```diff
@@ -0,0 +1,275 @@
...
+    file_sizes = dict(result)
+
+    if  args.group_by == 'sections':# or not args.group_by:
+        title = ['Sections','LoC']
+        results, irrelevant_dirs, relevant_dirs = \
+         get_counts(all_maintainers, file_sizes, filter_by_files, args.group_by)
+    elif args.group_by == 'maintainers':
+        title = ['Maintainers', 'LoC']
+        results, maintainer_to_section = \
+         get_counts(all_maintainers, file_sizes, filter_by_files, args.group_by)
```
rralf:
```
title = [args.group_by, 'LoC']
if args.group_by in {'sections', 'maintainers'}:
    result = get_counts(all_maintainers, file_sizes, filter_by_files, args.group_by)
```

And we get rid of the remaining redundancy.

bin/pasta_maintainers_stats.py
```diff
@@ -0,0 +1,275 @@
...
+                    counts[name].increase(loc, byte, filename in filter_by_files)
+
+    if len(filter_by_files) > 0:
+        # If filter is given, only return the relevant lines:
+        counts = {name:item for name,item in counts.items()
+                    if item.display_in_filter()}
+    if group_by == 'sections':
+        return decreasing_order(counts), filename_excluded, filenames_filt
+    else:
+        return decreasing_order(counts), maintainer_to_section
```
rralf: Mhhm. No... Returning different signatures shifts the problem to the caller-site. I wonder if we can somehow get a common datatype for both cases...

bin/pasta_maintainers_stats.py
```diff
@@ -0,0 +1,275 @@
...
+    results = []
+    for filename in filenames:
+        sections = all_maintainers.get_subsystems_by_file(filename)
+        sections -= {'THE REST'}
+        # To get a sense of how large the section is and order by it, look at
+        # how many different patterns it includes:
+        # all_maintainers.subsystems[section].files
+        sections = sorted(sections, 
+         key=lambda section: len(all_maintainers.subsystems[section].files))
+        results.append([filename, sections])
```
rralf: I don't get this ordering. What do you try to achieve? Sort subsystems by their size?

rralf (commented):
CC: Pia Eichinger @ditzlike 

Basak, Pia is currently working on improving LinuxMaintainers.py. Pia, could you please check that? I'll also have a look at it later.

rralf:
Basak, thanks for the PR. Will continue working on it and get back to you.

rralf:
@basakerdamar For your questions regarding the Maintainers...

Let's take v5.7-rc3 as reference (randomly chosen).

```
./scripts/get_maintainers.pl --subsystem -f ./Makefile
Masahiro Yamada <masahiroy@kernel.org> (maintainer:KERNEL BUILD + files below scripts/ (unless mai...)
Michal Marek <michal.lkml@markovi.net> (maintainer:KERNEL BUILD + files below scripts/ (unless mai...)
linux-kbuild@vger.kernel.org (open list:KERNEL BUILD + files below scripts/ (unless mai...)
linux-kernel@vger.kernel.org (open list)
clang-built-linux@googlegroups.com (open list:CLANG/LLVM BUILD SUPPORT)
KERNEL BUILD + files below scripts/ (unless maintained elsewhere)
THE REST
CLANG/LLVM BUILD SUPPORT
```

If I run in PaStA:

```
    maintainers_version = load_maintainers(config, {'v5.7-rc3'})
    maintainers = maintainers_version['v5.7-rc3']

    sections = maintainers.get_subsystems_by_file('Makefile')
    for section in sections:
        mtrs = maintainers.get_maintainers(section)
        print(mtrs)
```
It reports:
```
({'linux-kernel@vger.kernel.org'}, [('linus torvalds', 'torvalds@linux-foundation.org')], [])
({'linux-kbuild@vger.kernel.org'}, [('masahiro yamada', 'masahiroy@kernel.org'), ('michal marek', 'michal.lkml@markovi.net')], [])
```

which has two slight differences:
  - get_maintainers.pl will not show Linus Torvalds (unless --git-chief-penguins (yes, really) is specified)
  - PaStA misses the CLANG/LLVM BUILD SUPPORT section. Reason: LinuxMaintainers.py currently does not evaluate the K:-fields in MAINTAINERS.

So from my side, I would say everything looks fine.

Does that help?

rralf:
@ditzlike Note for us: PaStA lacks understanding for the K:-field.

ditzlike:
Noted. I'll see if I get to fix it this weekend

rralf:
Happy hacking. But this is going to be really challenging. For the moment we should probably skip the K-tags, and see how good our results are without them.

PR #56, opened 2020-05-14
Remove support for USE_PATCHWORK_ID

rsarky:
Made a fresh branch as lots of refactoring needs to be done. Things will be cleaner this way

### DIFF
--- a/bin/pasta_sync.py
+++ b/bin/pasta_sync.py
@@ -101,6 +101,5 @@ def sync(config, prog, argv):
         config.update_ccache_mbox()
 
         # Update the mail thread cache
-        if not config.mbox_use_patchwork_id:
-            repo.mbox.load_threads()
-            repo.mbox.threads.update()
+        repo.mbox.load_threads()
+        repo.mbox.threads.update()
--- a/pypasta/Repository/Mbox.py
+++ b/pypasta/Repository/Mbox.py
@@ -179,22 +179,19 @@ def load_file(filename, must_exist=True):
     return f
 
 
-def load_index(basename):
-    entries = load_file(basename, must_exist=False)
-    ret = dict()
-
-    for (date, message_id, location) in entries:
-        dtime = datetime.datetime.strptime(date, '%Y/%m/%d')
-
-        if message_id not in ret:
-            ret[message_id] = list()
-
-        ret[message_id].append((dtime, date, location))
-
-    return ret
-
-
 class MailContainer:
+    def load_index(self, f_index):
+        self.index = dict()
+        entries = load_file(f_index, must_exist=False)
+
+        for (date, message_id, location) in entries:
+            dtime = datetime.datetime.strptime(date, '%Y/%m/%d')
+
+            if message_id not in self.index:
+                self.index[message_id] = list()
+
+            self.index[message_id].append((dtime, date, location))
+
     def get_ids(self, time_window=None):
         if time_window:
             return {x[0] for x in self.index.items() if
@@ -217,7 +214,7 @@ class PubInbox(MailContainer):
         self.d_repo = d_repo
 
         self.repo = pygit2.Repository(d_repo)
-        self.index = load_index(self.f_index)
+        self.load_index(self.f_index)
 
         log.info('  ↪ loaded mail index for %s (shard %u): found %d mails' %
                  (listaddr, shard, len(self.index)))
@@ -321,7 +318,7 @@ class MboxRaw(MailContainer):
     def add_mbox(self, listname, f_mbox_raw):
         self.raw_mboxes.append((listname, f_mbox_raw))
         f_mbox_index = os.path.join(self.d_index, 'raw.%s' % listname)
-        index = load_index(f_mbox_index)
+        self.load_index(f_mbox_index)
         log.info('  ↪ loaded mail index for %s: found %d mails' % (listname, len(index)))
 
         for id, desc in index.items():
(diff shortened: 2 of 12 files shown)


bin/pasta_sync.py
```diff
@@ -101,6 +101,5 @@ def sync(config, prog, argv):
         config.update_ccache_mbox()
 
         # Update the mail thread cache
-        if not config.mbox_use_patchwork_id:
-            repo.mbox.load_threads()
-            repo.mbox.threads.update()
+        repo.mbox.load_threads()
+        repo.mbox.threads.update()
```
rralf, line 105: Hmm... We have to keep that spot in mind for the future: In case of a patchwork analysis, we will never need the mail thread structures. Those structures are typically used for analysis that go beyond mapping patches. So it's useless to load thread structures if we do a patchwork analysis.

But for the moment, your correction is okay!

pypasta/Repository/Mbox.py
```diff
@@ -184,7 +184,8 @@ def load_index(self, f_index):
         self.index = dict()
         entries = load_file(f_index, must_exist=False)
 
-        for (date, message_id, location) in entries:
+        for entry in entries:
+            date, message_id, location = (entry[0], entry[1], entry[2])
```
rralf, line 188: Unrelated change, will revert this.

pypasta/Repository/Mbox.py
```diff
@@ -266,7 +267,7 @@ def update(self, use_patchwork_id):
         hashes = hashes - known_hashes
         log.info('Updating %d emails' % len(hashes))
 
-        identifier = 'X-Patchwork-ID' if use_patchwork_id else 'Message-ID'
+        identifier = 'Message-ID'
```
rralf, line 270: We can further simplify this, will change it when applying the patch.

pypasta/Repository/Mbox.py
```diff
@@ -322,17 +320,11 @@ def add_mbox(self, listname, f_mbox_raw):
         self.raw_mboxes.append((listname, f_mbox_raw))
         f_mbox_index = os.path.join(self.d_index, 'raw.%s' % listname)
         self.load_index(f_mbox_index)
-        log.info('  ↪ loaded mail index for %s: found %d mails' % (listname, len(index)))
+        log.info('  ↪ loaded mail index for %s: found %d mails' % (listname, len(self.index)))
```
rralf, line 323: Oh, that's even a bug fix for a bug that I introduces earlier, right? Will take this to the bug-introducing patch and fix it there. Nice catch! Thanks!

rralf: Ah crap, I introduced a bigger problem with load_index, I need to reconsider it. I'll need to fix that first. I didn't respect that load_index will need to support sequential index loading for raw mboxes.

rralf (commented):
Hi Rohit,

thanks for the patch! I took a version this patch with slight changes to next. You pointed me towards a bug that I introduced with the load_index routine. Should be fixed on next now. However, I didn't yet test your patch. Especially, I need to test the shell parts. I'll leave you further comments on that soon. The rest looks good so far!

Thanks so far!

pypasta/Repository/Mbox.py
rsarky: This will be required, as if we have a patchwork id column, we will have 4 columns in an entry

pypasta/Repository/Mbox.py
rsarky: Sure, thanks

pypasta/Repository/Mbox.py
rsarky: Yup, realised that you must have missed that

pypasta/Repository/Mbox.py
rralf: Yes but not for the moment.

Even later, we can do it easier. E.g., sth. like:

for entry in entries:
  for field in entry
     ...

Then we're independant of the size of the tuple.

rralf:
Hi,

for the future: please go into deeper detail in your commit message. A reader should understand 'why' you exactly removed the USE_PATCHWORK_ID stuff. We have an explanation for that in our mail thread, but if someone looks after one year at that commit, he won't remind why we did it.

rsarky:
> Hi,
> 
> for the future: please go into deeper detail in your commit message. A reader should understand 'why' you exactly removed the USE_PATCHWORK_ID stuff. We have an explanation for that in our mail thread, but if someone looks after one year at that commit, he won't remind why we did it.

Yes, makes sense

pypasta/Repository/Mbox.py
rsarky: Hmm, for 3 or 4 items in a tuple looping through it seems like overkill

rralf: The alternative would be?

rsarky: >  date, message_id, location = (entry[0], entry[1], entry[2])

This?

rralf: And if there's a fourth entry in case of patchwork? We would - conditionally - have to maintain that list. We can avoid that with a simple generalisation.

rsarky: In any case the consumer needs to be aware of what columns of the index it needs. For 3-4 I find simple tuple destructuring better. If there are more columns than that, then yes, looping. Or maybe array slicing

rsarky:
Closing this as this has been merged to the `next` branch

rralf:
Yep, besides the commit "Add python-requests as dependency", as it's not required at the moment.

Thanks!

PR #57, opened 2020-05-17
Add patchwork interface

rsarky:
This PR adds supoort to pull patches from patchwork.
There are 2 ways PaStA can pull patches:
- A raw mailbox containing patches from patchwork. (can be used for initial import)
- Through Patchwork's API. In this case Pasta will only import patches that it doesnt have already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Requirements
 - python scikit-learn
 - python-toml
 - python-tqdm
+- python-requests
 - flask
   - flask-wtf
   - flask-bootstrap
--- a/README.md
+++ b/README.md
@@ -248,3 +248,46 @@ To compare all mails on the list against upstream:
 [1]: https://public-inbox.org/README.html
 [2]: https://github.com/xai/nntp2mbox
 [3]: https://git.kernel.org/pub/scm/public-inbox/
+
+PaStA and Patchwork
+--------------------
+The results of PaStA's analyses can be used by [Patchwork](http://jk.ozlabs.org/projects/patchwork/).
+
+## Setting up PaStA and Patchwork
+Assuming a working setup of PaStA already exists, here are the steps necessary for Patchwork integration
+
+1. Install Patchwork on your system, following the guidelines in Patchwork's [documentation](https://patchwork.readthedocs.io/en/latest/development/installation/)
+
+2. Start a shell inside Patchwork's docker container with `docker-compose run --rm web --shell`
+
+3. Bring up a Patchwork development server, by running
+`./manage.py runserver 0.0.0.0:8000` inside the shell started in step 2. You should now have a Patchwork instance
+running at `<Patchwork-Docker-Container-IP-address>:8000` on your host. The Patchwork container's IP address can be
+found using `ifconfig` command on Linux distributions.
+
+4. Start PaStA's docker container on the same network as the Patchwork one by running the command:
+`docker run -it --rm --network patchwork_default --name pasta -v </path/to/PaStA>:/home/pasta pasta:latest`
+
+5. Set the Patchwork specific settings in config:
+```
+[mbox]
+...
+
+[mbox.patchwork]
+url = 'http://<Patchwork-Container-IP-Address>/api/1.2/'
+projects = [{ id = x, initial_archive="path/to/archive", list_email="list@domain.org"},
+{id = y, initial_archive="", list_email="anotherlist@anotherdomain.org"}, ...]
+page_size = 10
+
+# Provide an api_token token or username/password if restricted api access is
+# needed (e.g updating relations)
+token = 'your_token'
+username = 'your_username'
+password = 'your_password'
+```
+
+Each Patchwork project from which mails are to be imported needs to be listed in the configuration. If
+the `initial_archive` property of the project is specified (project with id `x` in the example above)
+, PaStA will import mails from the archive treating it as a raw mail box. If the `initial_archive`
+property is an empty string (project with id `y` in the example above),
+PaStA will fetch mails using Patchwork's API, only importing those mails which are not already in PaStA.
--- a/pypasta/Repository/Mbox.py
+++ b/pypasta/Repository/Mbox.py
@@ -12,14 +12,17 @@ the COPYING file in the top-level directory.
 
 import email
 import glob
+import hashlib
 import os
 import pygit2
 import re
+import requests
 
 from datetime import datetime
 from email.charset import CHARSETS
 from logging import getLogger
 from subprocess import Popen
+from urllib.parse import urljoin
 
 from .MailThread import MailThread
 from .MessageDiff import MessageDiff, Signature
@@ -374,6 +377,116 @@ class MboxRaw(MailContainer):
         return ret
 
 
+class PatchworkProject(MailContainer):
+    NEXT_PAGE_REGEX = re.compile('.*[&?]page=(\d+).*')
+
+    def __init__(self, url, project_id, page_size, d_mbox, f_index, f_mbox_raw):
+        self.url = url
+        self.page_size = page_size
+        self.project_id = project_id
+        self.f_index = f_index
+        self.f_mbox_raw = f_mbox_raw
+        self.d_mbox = d_mbox
+        self.d_mbox_patchwork = os.path.join(self.d_mbox, 'patchwork')
+        self.index = self.load_index(self.f_index)
+        log.info('  ↪ loaded mail index for Patchwork project id %u: found %d mails' %
+                 (project_id, len(self.index)))
+
+
+    def _get_page(self, page):
+        params = dict()
+        params['order'] = 'id'
+        params['project'] = self.project_id
+        params['per_page'] = self.page_size
+        params['page'] = page
+
+        resp = requests.get(urljoin(self.url, 'patches'), params)
+        resp.raise_for_status()
+        json = resp.json()
+
+        page = {int(entry['id']):
+                    (datetime.fromisoformat(entry['date']),
+                     entry['msgid'],
+                     entry['mbox'])
+                for entry in json}
+
+        next_page = None
+        if 'next' in resp.links and 'url' in resp.links['next']:
+            next_url = resp.links['next']['url']
+            match = self.NEXT_PAGE_REGEX.match(next_url)
+            if match:
+                next_page = int(match.group(1))
+
+        return next_page, page
+
+    @staticmethod
+    def _pull_patch(url):
+        resp = requests.get(url)
+        resp.raise_for_status()
+        md5sum = hashlib.md5(resp.content).hexdigest()
+        return md5sum, resp.content
+
+    def update(self):
+        if self.f_mbox_raw:
+            log.info('Processing raw mailbox for Patchwork project id %u' %
+                     self.project_id)
+            process_mailbox_maildir(self.f_mbox_raw, str(self.project_id),
+                           self.d_mbox, mode='patchwork')
+            log.info('  ↪ Initial import successful. You can now remove the initial_archive in the configuration')
+            return
+
+        # Get a set of all present patchwork-ids
+        patchwork_ids = set() \
+            .union(*[{entry[3] for entry in idx_entry}
+                     for idx_entry in self.index.values()])
+        next_page = len(patchwork_ids)//self.page_size + 1
+        pulled = 0
+
+        try:
+            while next_page:
+                log.info('Querying page %u' % next_page)
+                next_page, page = self._get_page(next_page)
+
+                missing_ids = page.keys() - patchwork_ids
+                for missing_id in sorted(missing_ids):
+                    log.info(' Receiving patch %u' % missing_id)
+                    date, message_id, url = page[missing_id]
+                    format_date = date.strftime('%04Y/%m/%d')
+
+                    # Download the raw, unencoded patch
+                    md5sum, patch = self._pull_patch(url)
+
+                    # Persist it
+                    d_patch = os.path.join(self.d_mbox_patchwork, format_date)
+                    f_patch = os.path.join(d_patch, md5sum)
+                    os.makedirs(d_patch, exist_ok=True)
+                    if not os.path.exists(f_patch):
+                        with open(f_patch, 'wb') as f:
+                            f.write(patch)
+
+                    # Index it
+                    if message_id not in self.index:
+                        self.index[message_id] = list()
+                    self.index[message_id].append((date, format_date, md5sum, missing_id))
+                    pulled += 1
+        except Exception as e:
+            log.error(' An error occurred while fetching patches from Patchwork: %s' % str(e))
+
+        log.info('  ↪ Pulled %u patches in total' % pulled)
+        if pulled:
+            self.write_index(self.f_index)
+
+    def __getitem__(self, message_id):
+        ret = list()
+
+        for _, date_str, md5, _ in self.index[message_id]:
+            filename = os.path.join(self.d_mbox_patchwork, date_str, md5)
+            with open(filename, 'rb') as f:
+                ret.append(f.read())
+
+        return ret
+
+
 class Mbox:
     def __init__(self, config):
         self.threads = None
@@ -394,6 +507,21 @@ class Mbox:
             self.invalid |= {x[0] for x in load_file(f_inval)}
         log.info('  ↪ loaded invalid mail index: found %d invalid mails'
                  % len(self.invalid))
+
+        # If Patchwork projects are defined in the config, no need to load raw mboxes and pubins.
+        self.patchwork_projects = []
+        if len(config.patchwork['projects']):
+            log.info('Loading Patchwork projects...')
+        for project_id, f_mbox_raw, listaddr in config.patchwork['projects']:
+            self.lists.add(listaddr)
+            f_index = os.path.join(
+                config.d_mbox, 'index', 'patchwork.%u' % project_id)
+            project = PatchworkProject(config.patchwork['url'], project_id,
+                                       config.patchwork['page_size'],
+                                       self.d_mbox, f_index, f_mbox_raw)
+            for message_id in project.get_ids():
+                self.add_mail_to_list(message_id, listaddr)
+            self.patchwork_projects.append(project)
 
         if len(config.mbox_raw):
             log.info('Loading raw mailboxes...')
@@ -452,6 +580,10 @@ class Mbox:
         if message_id in self.mbox_raw:
             return True
 
+        for project in self.patchwork_projects:
+            if message_id in project:
+                return True
+
         return False
 
     def __getitem__(self, message_id):
@@ -478,6 +610,10 @@ class Mbox:
     def get_raws(self, message_id):
         raws = list()
 
+        for project in self.patchwork_projects:
+            if message_id in project:
+                raws += project[message_id]
+
         for public_inbox in self.pub_in:
             if message_id in public_inbox:
                 raws += public_inbox[message_id]
@@ -489,6 +625,9 @@ class Mbox:
 
     def get_ids(self, time_window=None, allow_invalid=False, lists=None):
         ids = set()
+
+        for project in self.patchwork_projects:
+            ids |= project.get_ids(time_window)
 
         for pub in self.pub_in:
             ids |= pub.get_ids(time_window)
@@ -504,6 +643,9 @@ class Mbox:
         return ids
 
     def update(self):
+        for project in self.patchwork_projects:
+            project.update()
+
         self.mbox_raw.update()
 
         for pub in self.pub_in:
--- a/docker/pasta-skeleton.dockerfile
+++ b/docker/pasta-skeleton.dockerfile
@@ -33,6 +33,7 @@ RUN apt install -y --no-install-recommends \
 	python3-toml \
 	python3-tqdm \
 	python3-wheel \
+	python3-requests \
 	sudo \
 	vim \
 	wget


rralf:
Thanks! I'll have a look at it tomorrow.

rralf:
One thing: Could you take the adjustments of Config.py and load/write_index to a separate commits?

rsarky:
> One thing: Could you take the adjustments of Config.py and load/write_index to a separate commits?

Sure

rsarky:
Updated

pypasta/Repository/Mbox.py
```diff
@@ -392,6 +492,20 @@ def __init__(self, config):
...
+            for project_id, f_mbox_raw, listaddr in config.mbox_patchwork_projects:
+                self.lists.add(listaddr)
+                f_index = os.path.join(
+                    config.d_mbox, 'index', 'patchwork.%u' % project_id)
+                project = PatchworkProject(
+                    project_id, self.d_mbox, f_index, f_mbox_raw, **config.patchwork)
+                for message_id in project.get_ids():
+                    self.add_mail_to_list(message_id, listaddr)
+                self.patchwork_projects.append(project)
+        else:
```
rralf: python complains about indentation here

rralf: I guess we can simply remove the else statement, right?

pypasta/Repository/Mbox.py
```diff
@@ -486,7 +605,10 @@ def get_raws(self, message_id):
 
     def get_ids(self, time_window=None, allow_invalid=False, lists=None):
         ids = set()
-
+        if len(self.patchwork_projects) > 0:
+            for project in self.patchwork_projects:
+                ids |= project.get_ids(time_window)
+        else:
```
rralf: same here, incorrect indentation

pypasta/Repository/Mbox.py
```diff
@@ -475,6 +589,11 @@ def get_messages(self, id):
     def get_raws(self, message_id):
         raws = list()
 
+        if len(self.patchwork_projects) > 0:
+            for project in self.patchwork_projects:
+                if message_id in project:
+                    raws += project[message_id]
+        else:
```
rralf: and here, my interpreter doesn't like it.

pypasta/Repository/Mbox.py
```diff
@@ -475,6 +589,11 @@ def get_messages(self, id):
     def get_raws(self, message_id):
         raws = list()
 
+        if len(self.patchwork_projects) > 0:
```
rralf: We don't need this check, the for loop will do it already for us

pypasta/Repository/Mbox.py
```diff
@@ -486,7 +605,10 @@ def get_raws(self, message_id):
 
     def get_ids(self, time_window=None, allow_invalid=False, lists=None):
         ids = set()
-
+        if len(self.patchwork_projects) > 0:
```
rralf: same here, the for loop will take care of this check.

pypasta/Repository/Mbox.py
```diff
@@ -501,6 +623,10 @@ def get_ids(self, time_window=None, allow_invalid=False, lists=None):
         return ids
 
     def update(self):
+        if len(self.patchwork_projects) > 0:
+            for project in self.patchwork_projects:
+                project.update()
+        else:
```
rralf: indentation.

pypasta/Repository/Mbox.py
```diff
@@ -501,6 +623,10 @@ def get_ids(self, time_window=None, allow_invalid=False, lists=None):
         return ids
 
     def update(self):
+        if len(self.patchwork_projects) > 0:
```
rralf: remove the check.

rralf (changes requested):
Hi,

before I continue with my review: please remove unnecessary if len() checks, and take care of the indentation. In all cases, you can remove the else statement. While it's surely not a good idea, there's no need to disallow raw mboxes in combination with patchwork. It's not required for our use-case, but there's no reason to forbid it.

rralf (changes requested):
This patch breaks PaStA if we don't use a patchwork configuration:

  File "/media/Datenfass/PaStA/pypasta/Config.py", line 199, in __init__
    mbox_patchwork = mbox['patchwork']
KeyError: 'patchwork'

Please take care to not break the system. Only address patchwork-related config items, if a patchwork configuration is being used.

rralf (changes requested):
Please split it up into two patches: The first one to change load_index, the second one that pulls out the method.

pypasta/Repository/Mbox.py
```diff
@@ -179,28 +179,51 @@ def load_file(filename, must_exist=True):
...
+    if ret == 0:
+        log.info('  ↪ done')
+    else:
+        log.error('Mail processor failed!')
+
+
 class MailContainer:
     @staticmethod
-    def load_index(f_index):
+    def load_index(f_index, is_patchwork_index=False):
```
rralf: TBH, I don't like this approach. We're calling the method like

load_index('/filename', True)

No one knows what "True" or "False" means on caller-site, if we use it as an argument. And I'm sure we don't even need this argument.

pypasta/Repository/Mbox.py
```diff
@@ -179,28 +179,51 @@ def load_file(filename, must_exist=True):
...
+            dtime = datetime.strptime(date, '%Y/%m/%d')
 
             if message_id not in index:
                 index[message_id] = list()
 
-            index[message_id].append((dtime, date, location))
+            if is_patchwork_index:
+                index[message_id].append((dtime, date, location, entry[3]))
+            else:
+                index[message_id].append((dtime, date, location))
```
rralf: One can only guess what entry[3] means, and know we have a rather complex differentiation, depending on is_patchwork_index.

rralf: Let me propose a different approach, will get back to you.

pypasta/Repository/Mbox.py
rsarky: Since it's a keyword argument, the caller will have to include the keyword ie `is_patchwork_index` while making the call. Thus making it explicit.

rsarky:
> Hi,
> 
> before I continue with my review: please remove unnecessary if len() checks, and take care of the indentation. In all cases, you can remove the else statement. While it's surely not a good idea, there's no need to disallow raw mboxes in combination with patchwork. It's not required for our use-case, but there's no reason to forbid it.

In all the cases, the else and the conditions based on len were introduced to disallow PaStA using Patchwork and public or raw mailboxes simultaneously. If we do want them to function together, I guess I can remove those checks

rralf:
> > Hi,
> > before I continue with my review: please remove unnecessary if len() checks, and take care of the indentation. In all cases, you can remove the else statement. While it's surely not a good idea, there's no need to disallow raw mboxes in combination with patchwork. It's not required for our use-case, but there's no reason to forbid it.
> 
> In all the cases, the else and the conditions based on len were introduced to disallow PaStA using Patchwork and public or raw mailboxes simultaneously. If we do want them to function together, I guess I can remove those checks

Yep, let's remove them.

rsarky:
> This patch breaks PaStA if we don't use a patchwork configuration:
> 
> File "/media/Datenfass/PaStA/pypasta/Config.py", line 199, in **init**
> mbox_patchwork = mbox['patchwork']
> KeyError: 'patchwork'
> 
> Please take care to not break the system. Only address patchwork-related config items, if a patchwork configuration is being used.

The same thing would happen if we removed `[mbox.raw]` or `[mbox.pubin]` from the Config. This PR assumes that the config will by default contain the line `[mbox.patchwork]`, this is the same way as pubins or raw mboxes are handled

rralf:
> > This patch breaks PaStA if we don't use a patchwork configuration:
> > File "/media/Datenfass/PaStA/pypasta/Config.py", line 199, in **init**
> > mbox_patchwork = mbox['patchwork']
> > KeyError: 'patchwork'
> > Please take care to not break the system. Only address patchwork-related config items, if a patchwork configuration is being used.
> 
> The same thing would happen if we removed `[mbox.raw]` or `[mbox.pubin]` from the Config. This PR assumes that the config will by default contain the line `[mbox.patchwork]`, this is the same way as pubins or raw mboxes are handled

Oh, does it? Thanks for the bug report! :+1: 
Will fix this...

rsarky:
> Please split it up into two patches: The first one to change load_index, the second one that pulls out the method.


I didnt pull out the `load_index` method. That was done by you in a previous commit.
Are you talking about the `load_mbox_file` method by any chance?

rsarky:
I wonder why the indentation is breaking. I used to use vim, but now I switched to vscode that uses pep8 style for formatting

rsarky:
> Hi,
> 
> before I continue with my review: please remove unnecessary if len() checks, and take care of the indentation. In all cases, you can remove the else statement. While it's surely not a good idea, there's no need to disallow raw mboxes in combination with patchwork. It's not required for our use-case, but there's no reason to forbid it.

Done

pypasta/Repository/Mbox.py
rralf: Please have a look at the latest two commits on next:
https://github.com/lfd/PaStA/commit/fc11490ef6ab5a7795c2d9d3d3ead12be8a892db and https://github.com/lfd/PaStA/commit/ae26e9be8713da08311ad9040bb8a23a65b3794e .

Should work with public inboxes as well as the patchwork backend. The patchwork-id will only be loaded/written if present. What do you think?

rralf:
> > > This patch breaks PaStA if we don't use a patchwork configuration:
> > > File "/media/Datenfass/PaStA/pypasta/Config.py", line 199, in **init**
> > > mbox_patchwork = mbox['patchwork']
> > > KeyError: 'patchwork'
> > > Please take care to not break the system. Only address patchwork-related config items, if a patchwork configuration is being used.
> > 
> > 
> > The same thing would happen if we removed `[mbox.raw]` or `[mbox.pubin]` from the Config. This PR assumes that the config will by default contain the line `[mbox.patchwork]`, this is the same way as pubins or raw mboxes are handled
> 
> Oh, does it? Thanks for the bug report! +1
> Will fix this...

Aah, we do not have a bug here. We set default values in resource's common.cfg:

https://github.com/lfd/PaStA-resources/blob/master/common/default.cfg

So the fix is simple: In future, we must add an empty [mbox.patchwork] entry to default.cfg. But we can postpone that for the moment.

rsarky:
> > > > This patch breaks PaStA if we don't use a patchwork configuration:
> > > > File "/media/Datenfass/PaStA/pypasta/Config.py", line 199, in **init**
> > > > mbox_patchwork = mbox['patchwork']
> > > > KeyError: 'patchwork'
> > > > Please take care to not break the system. Only address patchwork-related config items, if a patchwork configuration is being used.
> > > 
> > > 
> > > The same thing would happen if we removed `[mbox.raw]` or `[mbox.pubin]` from the Config. This PR assumes that the config will by default contain the line `[mbox.patchwork]`, this is the same way as pubins or raw mboxes are handled
> > 
> > 
> > Oh, does it? Thanks for the bug report! +1
> > Will fix this...
> 
> Aah, we do not have a bug here. We set default values in resource's common.cfg:
> 
> https://github.com/lfd/PaStA-resources/blob/master/common/default.cfg
> 
> So the fix is simple: In future, we must add an empty [mbox.patchwork] entry to default.cfg. But we can postpone that for the moment.

Yup makes sense

rralf:
Rebased your work and pushed it to wip-patchwork-integration. Will test it now.

README.md
```diff
@@ -248,3 +248,47 @@ To compare all mails on the list against upstream:
...
+
+4. Start PaStA's docker container on the same network as the Patchwork one by running the command:
+`docker run -it --rm --network patchwork_default --name pasta -v </path/to/PaStA>:/home/pasta pasta:latest`
+
+5. Set the Patchwork specific settings in config:
+```
+[mbox]
+...
+
+USE_PATCHWORK_ID = true
```
rralf: no longer required

pypasta/Repository/Mbox.py
```diff
@@ -392,6 +492,20 @@ def __init__(self, config):
         log.info('  ↪ loaded invalid mail index: found %d invalid mails'
                  % len(self.invalid))
 
+        # If Patchwork projects are defined in the config, no need to load raw mboxes and pubins.
+        self.patchwork_projects = []
+        if len(config.mbox_patchwork_projects):
+            log.info('Loading Patchwork projects...')
+            for project_id, f_mbox_raw, listaddr in config.mbox_patchwork_projects:
```
rralf: The loop can go down one indentation level

pypasta/Repository/Mbox.py
```diff
@@ -392,6 +492,20 @@ def __init__(self, config):
...
+                self.lists.add(listaddr)
+                f_index = os.path.join(
+                    config.d_mbox, 'index', 'patchwork.%u' % project_id)
```
rralf: Yep, looks good!

pypasta/Repository/Mbox.py
```diff
@@ -392,6 +492,20 @@ def __init__(self, config):
...
+                project = PatchworkProject(
+                    project_id, self.d_mbox, f_index, f_mbox_raw, **config.patchwork)
+                for message_id in project.get_ids():
+                    self.add_mail_to_list(message_id, listaddr)
+                self.patchwork_projects.append(project)
+
```
rralf, line 525: Hmm. I can see a lot of duplicated startup code for all three different mailboxes. Maybe we can somehow consolidate the startup. But let's do that later.

pypasta/Repository/Mbox.py
```diff
@@ -371,6 +375,102 @@ def __getitem__(self, message_id):
...
+        log.info('Downloading patches from Patchwork for project id %u' % self.project_id)
+        params = {}
+        params['order'] = '-id'   # descending id order
+        params['project'] = self.project_id
+        resp = requests.get(urljoin(self.patchwork_url, 'patches'), params)
+        return self.page_iterator(resp, yield_value, since)
+
+    def update(self):
+        if self.f_mbox_raw != "":
```
rralf: Can be simplified to:
if self.f_mbox_raw:

pypasta/Repository/Mbox.py
```diff
@@ -16,10 +16,14 @@
...
 import pygit2
 import re
+import requests
+import hashlib
 
 from email.charset import CHARSETS
 from logging import getLogger
 from subprocess import call
+from urllib.parse import urljoin
+from datetime import datetime
```
rralf: Hmm, above, we already import datetime. Importing it this way requires to maintain existing code.

pypasta/Repository/Mbox.py
```diff
@@ -371,6 +375,102 @@ def __getitem__(self, message_id):
...
+        params['project'] = self.project_id
+        resp = requests.get(urljoin(self.patchwork_url, 'patches'), params)
+        return self.page_iterator(resp, yield_value, since)
+
+    def update(self):
+        if self.f_mbox_raw != "":
+            log.info('Processing raw mailbox for Patchwork project id %u' %
+                     self.project_id)
+            load_mbox_file(self.f_mbox_raw, str(self.project_id),
+                           self.d_mbox, mode='patchwork')
```
rralf, line 434: Add a report here that the initial import was successful, and guide the user to remove the initial archive configuration option. Return here.

pypasta/Repository/Mbox.py
```diff
@@ -371,6 +375,102 @@ def __getitem__(self, message_id):
...
+        else:
```
rralf: And remove the else case, and put the indentation one level down.

pypasta/Repository/Mbox.py
```diff
@@ -371,6 +375,102 @@ def __getitem__(self, message_id):
...
+            greatest_patchwork_id = None
+            if len(self.index) > 0:
+                greatest_patchwork_id = max(
+                    [entry[3] for entries in self.index.values() for entry in entries])
```
rralf: Shouldn't we start from greatest_patchwork_id + 1? Otherwise, we will create duplicate entries for the highest patchwork id.

rralf (changes requested):
Hi,

just successfully run your script. There are still some issues, we receive duplicate messages. Let's talk about that in a short video conference. Will send you an email tomorrow.

Thanks
  Ralf

pypasta/Repository/Mbox.py
```diff
@@ -371,6 +375,102 @@ def __getitem__(self, message_id):
...
+            patches = self.download_patches(since=greatest_patchwork_id)
+            pulled = 0
+            for patch in patches:
+                date, message_id, message_text, patchwork_id = patch
```
rralf: Hmm. We need to either get the API under control, or check if the patchwork-id is already present.

pypasta/Repository/Mbox.py
rsarky: No. You can check the code in the `page_iterator` method, we "break" out of the generator when we encounter the patch with the highest patch id without yielding it.

pypasta/Repository/Mbox.py
rsarky: We can do some empirical tests to see if the API returns a patch with the same patchwork id twice, but I dont think this should happen. I am not sure if this is what you mean

pypasta/Repository/Mbox.py
rsarky: Yes, that's a good idea

README.md
rsarky: > We should allow to have no initial_archive set in the configuration file.

So do you mean instead of checking for empty string in `initial_archive` property, we let the user completely remove that key?

It seems like the another way to do the same thing. Are there any advantages to remove the key instead of setting it to an empty string?

rsarky:
> Hi,
> 
> just successfully run your script. There are still some issues, we receive duplicate messages. Let's talk about that in a short video conference. Will send you an email tomorrow.
> 
> Thanks
> Ralf

Sure

README.md
rralf: Usually people comment out configuration options that they don't need. This is what I intuitively tried. I commented out initial_archive after the first run. Resulted in a crash of PaStA.

rsarky: initial_archive key is associated with a projectid and list email. ie it is part of the same object.
So I dont think that can be commented out.

If we keep a separate array of initial_archives and separate array of projects that might be troublesome as both the arrays may have conflicting information

pypasta/Repository/Mbox.py
```diff
@@ -374,6 +377,100 @@ def __getitem__(self, message_id):
...
+                     self.project_id)
+            process_mailbox_maildir(self.f_mbox_raw, str(self.project_id),
+                           self.d_mbox, mode='patchwork')
+            log.info('  ↪ Initial import successful. You can now remove the initial_archive in the configuration')
+            return
+
+        pages = self.download_patches()
+        pulled = 0
+        for index, page in enumerate(pages):
+            for patch in page:
```
rralf: This loop implicitly calls the iterator. The iterator does IO, and can fail. Please don't combine iterators with side-effects and IO.

rralf: Other than that, we download all patches of the page, which is not necessary, as we should only load the ones that we're missing.

rralf: Will return with a patch that implements that.

rralf (changes requested):

rsarky:
Added some error handling. Tested it out by stopping the patchwork server while the update was running. Works!

Added it to [this](https://github.com/lfd/PaStA/pull/57/commits/bd498e60542286cbe028f85fad3cbed702c98e60) commit

pypasta/Repository/Mbox.py
```diff
@@ -374,6 +377,122 @@ def __getitem__(self, message_id):
...
+        # Get a set of all present patchwork-ids
+        patchwork_ids = set() \
+            .union(*[{entry[3] for entry in idx_entry}
+                     for idx_entry in self.index.values()])
+        next_page = len(patchwork_ids)//self.page_size + 1
+        pulled = 0
+        error = False
+        while next_page and not error:
+            log.info('Querying page %u' % next_page)
+            try:
```
rralf: We can do it simpler: Use one big try block for the while loop. It doesn't matter if the error occured in _get_page or _pull_patch. We have to stop in any case.

pypasta/Repository/Mbox.py
```diff
@@ -374,6 +377,122 @@ def __getitem__(self, message_id):
...
+            log.info('  ↪ Initial import successful. You can now remove the initial_archive in the configuration')
+            return
+
+        # Get a set of all present patchwork-ids
+        patchwork_ids = set() \
+            .union(*[{entry[3] for entry in idx_entry}
+                     for idx_entry in self.index.values()])
+        next_page = len(patchwork_ids)//self.page_size + 1
+        pulled = 0
+        error = False
```
rralf: try:
[...]

pypasta/Repository/Mbox.py
```diff
@@ -374,6 +377,122 @@ def __getitem__(self, message_id):
...
+        while next_page and not error:
+            log.info('Querying page %u' % next_page)
+            try:
+                next_page, page = self._get_page(next_page)
+            except requests.exceptions.RequestException:
```
rralf: That's not sufficient. I just got:
urllib3.exceptions.ProtocolError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))

RequestException is only one possible error. We need wider exception handling.

pypasta/Repository/Mbox.py
```diff
@@ -374,6 +377,122 @@ def __getitem__(self, message_id):
...
+                log.info(' Error fetching page %u. Stopping update' % next_page)
```
rralf: log.error, as it's an error.

No need to print next_page, we just printed it in line 446. Better print the actual cause of the eception

catch Exception as e:
  log.error(' An error occured during communication with patchwork:')
  log.error(' %s' % str(e))

rralf (changes requested):

pypasta/Repository/Mbox.py
rsarky: Hmm, makes sense

pypasta/Repository/Mbox.py
rsarky: Seems like that was a problem with the url, not the http requests.
Anyway, seems like I will introduce a "catch all" then

rralf:
Looks good now! Thanks, merged to next and wip-patchwork-integration.

PR #58, opened 2020-06-06
Form patchwork relations

rsarky:


pypasta/Repository/Mbox.py
```diff
@@ -656,6 +659,13 @@ def update(self):
     def get_lists(self, message_id):
         return self.message_id_to_lists[message_id]
 
+    def get_patchwork_ids(self, message_id):
+        ret = list()
```
rralf: set()

pypasta/Repository/Mbox.py
```diff
@@ -656,6 +659,13 @@ def update(self):
...
+        for project in self.patchwork_projects:
+            if message_id in project:
+                ret.append(project.get_patchwork_id(message_id))
```
rralf: ret |= project.get_patchwork_ids(message_id)

pypasta/Repository/Mbox.py
```diff
@@ -488,6 +488,9 @@ def __getitem__(self, message_id):
 
         return ret
 
+    def get_patchwork_id(self, message_id):
```
rralf, line 491: must be get_patchwork_ids

bin/pasta_form_patchwork_relations.py
```diff
@@ -0,0 +1,62 @@
+import sys
```
rralf: Missing copyright header

bin/pasta_form_patchwork_relations.py
```diff
@@ -0,0 +1,62 @@
...
+import os
```
rralf, line 11: Please order import statements alphabetically

bin/pasta_form_patchwork_relations.py
```diff
@@ -0,0 +1,62 @@
...
+
+    config.repo.register_mbox(config)
+    mbox = config.repo.mbox
+
+    if args.infile is None:
+        f_clustering = config.f_clustering
+    else:
+        f_clustering = args.infile
+
+    f_result = 'patch-groups-patchwork'
```
rralf: This should be inside resources/project/resources. You can add the filename to the default option of the argument parser.

bin/pasta_form_patchwork_relations.py
```diff
@@ -0,0 +1,62 @@
...
+    args = parser.parse_args(argv)
+
+    if config.mode != config.Mode.MBOX:
+        log.error('Only works in Mbox mode!')
+        return -1
+
+    config.repo.register_mbox(config)
+    mbox = config.repo.mbox
+
+    if args.infile is None:
```
rralf: If args.infile takes a sane default value, then this is superfluous.

pasta
```diff
@@ -83,6 +84,7 @@ def usage(me, exit_code=errno.EINVAL):
           '  ripup\n'
           '  upstream_history\n'
           '  web\n'
+          '  form_patchwork_relations\n'
```
rralf: Please keep the alphabetical order.

pasta
```diff
@@ -169,6 +171,8 @@ def main(argv):
         return upstream_duration(config, sub, argv)
     elif sub == 'web':
         return web(config, sub, argv)
+    elif sub == 'form_patchwork_relations':
+        return form_patchwork_relations(config, sub, argv)
```
rralf: same here.

bin/pasta_form_patchwork_relations.py
```diff
@@ -0,0 +1,62 @@
...
+            patch = mbox.get_patchwork_ids(item)
+            new_line += patch
+            count += 1
+        line[:count] = new_line
+
+    with open(f_result, 'w+') as output:
+        result = list()
+        for line in f:
+            result.append(' '.join([str(x) for x in line]))
+        output.write('\n'.join(result) + '\n')
```
rralf: No, please don't do it this way. We have an API for this kind of stuff. You're reintroducing the patch-groups parser which we already have inside pypasta/Clustering.py.

What you should rather do is something like:

```
clustering = config.load_cluster(...)
clustering_patchwork = Clustering()

for downstream, upstream in clustering.iter_split():
    patchwork_ids = {mbox.get_patchwork_ids(p) for p in downstream)
    clustering_patchwork.insert(patchwork_ids)
    ...

clustering_patchwork.to_file(args.infile)
```

rralf (changes requested):
See comments

rralf (changes requested):
see comments

pypasta/Repository/Mbox.py
rsarky: For a particular Patchwork project, a message id will always map to a single patchwork id

bin/pasta_form_patchwork_relations.py
rsarky: Yes, I was a bit confused as to the content of the header. Should I just replace your name with mine in the other headers. Do I also need to change `OTH Regensburg`, if yes to what? The name of my college?

bin/pasta_form_patchwork_relations.py
rsarky: sure

bin/pasta_form_patchwork_relations.py
rsarky: yes, makes sense

bin/pasta_form_patchwork_relations.py
rsarky: I guess we should also look for the `infile` in /resources/project/resources to be consistent then

rsarky: Otherwise what if for both infile and outfile the user can define whatever path he wants explicitly. If he doesnt define either of those two, we pick up the appropriate default from /resources/project/resources

pasta
rsarky: sure

pasta
rsarky: sure

bin/pasta_form_patchwork_relations.py
rsarky: Yes, using this API makes more sense!

rsarky:
Do we reallly need to include upstreams in the generated file?
I dont see Patchwork using those upstream references as they have a different model.
Just listing out patch groups looks cleaner.
What do you think?

pypasta/Repository/Mbox.py
rralf: What if two different messages with the same message id will be sent to a list? However, there is a chance that the case could happen, and we wouldn't treat it as we should.

bin/pasta_form_patchwork_relations.py
rralf: You don't do it on behalf of your college. In your case, i'd say to simply use your own name as copyright holder.

bin/pasta_form_patchwork_relations.py
rralf: Yes, we can make it an optional argument. But I'd say that it probably won't happen that the user wants to specify a specific patch-groups file other than the one in resources/...

rralf:
With upstreams, do you mean all upstream commit hashes, or just the ones without a mapping?

rsarky:
> With upstreams, do you mean all upstream commit hashes, or just the ones without a mapping?

All. Patchworks model currently maps a patch to a single upstream reference. That will be difficult for us to use.

pypasta/Repository/Mbox.py
rsarky: Afaik (project_id, message_id) is guaranteed to be unique in patchwork. They ignore subsequent messages to the same project with the same message_id iirc

bin/pasta_form_patchwork_relations.py
rsarky: Yup, was thinking in terms of enforcing mechanism and not policy :)

rralf:
Yikes. But eventually, we do want to point to upstream commit, so we should not simply omit them. So for patches that only have one upstream commit hash (i.e., most cases), we have no problem. We only have a problem for patch groups with more than one upstream commit hash.

For these cases, we somehow need to decide which commit hash to choose. In that case, I'd go for the oldest one.

rsarky:
Almost everytime that we have a patch group that consists of a patch set we will have multiple upstream commit references. In this case wouldn't it be misleading for each patch to point to the same upstream commit? Ideally each should be mapped to a unique reference

rsarky:
Also, there are users of this commit reference in Patchwork. [example](https://lore.kernel.org/patchwork/patch/1218385/). I am not sure if this was manually done or there' some logic in Patchwork that does this.

rralf:
I don't know either. @bulwahn Any ideas where upstream commit hashes in lore's instances do come from?

For the rest: Yes, in the first run, we can leave out upstream commit hashes to get things forward.

rralf:
Much nicer! Thanks, merged to next, together with another commit that add support for integer in Clusterings. In this way, we can return integers from get_patchwork_ids. In addition to that, I changed get_patchwork_id to get_patchwork_ids... I somehow feel better with that change. I would just like to ask you for on additional thing:

Could you please add a default configuration variable for the default destination location of the output in the default.cfg in the resources repository and use it? Thanks!

PR #59, opened 2020-06-08
Refactoring aggregation analyses

anmolsingh20:
Hi Ralf,

Here is a refactored version of the pasta evaluation part. The off-list code is not integrated yet, but a placeholder is provided. The analyses will follow in a separate pull-request.

Thanks

### DIFF
--- a/bin/pasta_evaluate_patches.py
+++ b/bin/pasta_evaluate_patches.py
@@ -7,6 +7,7 @@ Copyright (c) OTH Regensburg, 2019-2020
 Authors:
   Sebastian Duda <sebastian.duda@fau.de>
   Ralf Ramsauer <ralf.ramsauer@oth-regensburg.de>
+  Anmol Singh <anmol.singh@bmw.de>
 
 This work is licensed under the terms of the GNU GPL, version 2. See
 the COPYING file in the top-level directory.
@@ -22,7 +23,7 @@ from subprocess import call
 from pypasta.LinuxMaintainers import load_maintainers
 from pypasta.LinuxMailCharacteristics import load_linux_mail_characteristics
 
-log = getLogger(__name__[-15:])
+log = getLogger(__name__[-16:])
 
 _repo = None
 _config = None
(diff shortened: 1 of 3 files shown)


pasta
```diff
@@ -147,8 +147,8 @@ def main(argv):
         return check_mbox(config, sub, argv)
     elif sub == 'compare':
         return compare(config, sub, argv)
-    elif sub == 'evaluate_patches':
-        return evaluate_patches(config, sub, argv)
+    elif sub == 'prepare_evaluation':
+        return prepare_evaluation(config, sub, argv)
```
rralf, line 151: Misses the update in the usage(). Will add it when merging.

bin/pasta_prepare_evaluation.py
```diff
@@ -15,15 +15,21 @@
...
 from subprocess import call
 
-from pypasta.LinuxMaintainers import load_maintainers
+from tqdm import tqdm
+
+from pypasta.LinuxMaintainers import LinuxMaintainers
 from pypasta.LinuxMailCharacteristics import load_linux_mail_characteristics
 
-log = getLogger(__name__[-16:])
+log = getLogger(__name__[-18:])
```
rralf, line 32: Is there a particular reason for this change?

bin/pasta_prepare_evaluation.py
```diff
@@ -15,15 +15,21 @@
 
 import argparse
 import csv
+import os
 import re
+import pickle
```
rralf, line 20: Please maintain alphabetical order. Will change this when merging.

bin/pasta_evaluate_patches.py
```diff
@@ -316,8 +317,55 @@ def prepare_off_list_patches():
...
+            # Handle entries with patches
+            cluster_dict = {'cluster_id': clustering.get_cluster(patch_id), 'patch_id': patch_id, 'upstream': u}
+
+            # Add responses for the patch
+            response_lst = []
+            _load_responses_dict(patch_id, response_lst)
+            cluster_dict['responses'] = response_lst
+            clusters_responses.append(cluster_dict)
+
+    with open('patch_responses.pickle', 'wb') as handle:
```
rralf: Please use a filename that is provided through the configuration and store the pickle file in the resources directory structure.

rralf:
Hi Anmol,

merged it with minor modifications to next, thanks. On top of the series, could you please provide a patch that uses a standardised location for your pickle files?

Thanks

anmolsingh20:
> Hi Anmol,
> 
> merged it with minor modifications to next, thanks. On top of the series, could you please provide a patch that uses a standardised location for your pickle files?
> 
> Thanks

Ok, will do this in the next patch. thanks

bin/pasta_prepare_evaluation.py
anmolsingh20: The log info doesn't print the whole name of the function otherwise. Length of 'prepare_evaluation' is 18 here. I do not like this format of logging in general, but I just complied with the rest of the logger code.

rralf: Heh, no worries, I don't like it either! :-)
Need to reconsider the whole logging structure one day...

PR #60, opened 2020-06-18
Refactoring aggregation analyses

anmolsingh20:
Hi Ralf, could you please have a look. I will probably need to rebase again, but it is a moving target, so will do after a review. There is an additional pull request on the resources repo to complete this change.


rralf:
Hi Anmol,
I rebased to next, and accepted both patches. For the second one, I referenced the new commit hash of the resources submodule, and removed all changes that are not related to the change of the location of the configuration. The reason: I don't understand why you import pandas, flat_table or dask. There are currently no users of those libraries.

anmolsingh20:
> Hi Anmol,
> I rebased to next, and accepted both patches. For the second one, I referenced the new commit hash of the resources submodule, and removed all changes that are not related to the change of the location of the configuration. The reason: I don't understand why you import pandas, flat_table or dask. There are currently no users of those libraries.

I missed removing those imports while rolling back some other changes, so it is good that you removed them. They will follow in the next patch. Thanks!

PR #61, opened 2020-06-25
Extend pasta prepare_evaluation to handle review analysis

anmolsingh20:
Hi Ralf,

Here is a concise script form of the first basic filtering and analysis of patch review. I could rebase it on the newest next once the review is through.

Many thanks,
Anmol

### DIFF
--- a/analyses/response_analysis.py
+++ b/analyses/response_analysis.py
@@ -1,0 +1,151 @@
+
+
+import dask.dataframe as dd
+import matplotlib.pyplot as plt
+import pandas as pd
+import seaborn as sns
+
+from logging import getLogger
+log = getLogger(__name__)
+
+
+def analyse_responses(responses_csv):
+
+    def _process_patch_ids(patch_ids):
+        patch_id_list = list(patch_ids)
+        try:
+            patch_id_list.remove('_')
+        except ValueError:
+            pass
+        return len(set(patch_id_list))
+
+    final = dd.read_csv(responses_csv, blocksize=50e7,
+                        dtype={"idx ": "int32", "patch_id ": "category", \
+                               "responses.resp_msg_id": "category", \
+                               "responses.parent": "category", \
+                               "upstream": "category", \
+                               "response_author": "category"}).drop('Unnamed: 0', axis=1)
+
+    author_patch_counts_dask = final[['patch_id', 'response_author']].groupby('response_author')['patch_id'].agg(
+        'count')
+
+    author_patch_counts_dask.nlargest(20).compute(). \
+        plot(kind='barh', stacked=False, figsize=[10, 8], colormap='hsv')
+    plt.title('Top 20 responders by patches reviewed')
+    plt.ylabel('Responding authors')
+    plt.xlabel('Number of patch responses ')
+    plt.tight_layout()
+    plt.savefig('author_top20_patch_counts.pdf')
+    plt.close()
+
+    upstream_patch_counts_dask = final[['upstream', 'patch_id']].groupby('upstream')['patch_id'].apply(
+        _process_patch_ids, meta=pd.Series([], dtype=int))
+
+    log.info("Unique patch ids {}".format(final['patch_id'].nunique().compute(num_workers=20)))
+
+    log.info("Unique upstream commits {}".format(final['upstream'].nunique().compute(num_workers=20)))
+
+    upstream_patch_counts_dask.nlargest(10).compute(). \
+        plot(kind='barh', stacked=False, figsize=[10, 5], colormap='hsv')
+    plt.title('Top 10 upstreams by number of related patches')
+    plt.ylabel('Upstream commits')
+    plt.xlabel('Number of patches')
+    plt.tight_layout()
+    plt.savefig('upstream_top10_patch_counts.pdf')
+    plt.close()
+
+    plt.figure(figsize=(10, 7))
+    ax = sns.distplot(upstream_patch_counts_dask.compute(), bins=40, kde=False, color='blue', \
+                      vertical=False, kde_kws={"clip": (0, 40)}, hist_kws={"range": (0, 40)})
+    ax.set_yscale('log')
+    plt.title('Upstream Distribution over patches')
+    plt.xlabel('patch count')
+    plt.ylabel('upstream')
+    plt.tight_layout()
+    plt.savefig('upstream_dist_patch_counts_40bins.pdf')
+    plt.close()
+
+    upstream_response_counts_dask = final[['upstream', 'responses.resp_msg_id']].groupby('upstream')[
+        'responses.resp_msg_id'].agg('count')
+
+    upstream_response_counts_dask.nlargest(15).compute(). \
+        plot(kind='barh', stacked=False, figsize=[10, 6], colormap='Pastel1')
+    plt.title('Top 15 upstream commits by number of responses')
+    plt.ylabel('Upstream commits')
+    plt.xlabel('Number of responses')
+    plt.tight_layout()
+    plt.savefig('upstream_top15_response_counts.pdf')
+    plt.close()
+
+    plt.figure(figsize=(10, 8))
+    ax = sns.distplot(upstream_response_counts_dask.compute(), bins=500, kde=False, color='green',
+                      vertical=False, kde_kws={"clip": (0, 500)}, hist_kws={"range": (0, 500)})
+    ax.set_yscale('log')
+    plt.title('Upstream Distribution over number of responses')
+    plt.xlabel('response count')
+    plt.ylabel('upstream')
+    plt.tight_layout()
+    plt.savefig('upstream_dist_response_counts_500bins.pdf')
+    plt.close()
+
+    plt.figure(figsize=(10, 8))
+    ax = sns.distplot(upstream_response_counts_dask.compute(), bins=200, kde=False, color='blue',
+                      vertical=False, kde_kws={"clip": (0, 200)}, hist_kws={"range": (0, 200)})
+    ax.set_yscale('log')
+    plt.title('Upstream Distribution over responses, 200 bins')
+    plt.xlabel('responses')
+    plt.ylabel('upstream count')
+    plt.tight_layout()
+    plt.savefig('upstream_dist_response_counts_200bins.pdf')
+    plt.close()
+
+    patch_id_response_counts_dask = final[['patch_id', 'responses.resp_msg_id']].groupby('patch_id')[
+        'responses.resp_msg_id'].agg('count')
+
+    plt.figure(figsize=(12, 12))
+    ax = sns.distplot(patch_id_response_counts_dask.compute(), bins=500, kde=False, color='orange',
+                      vertical=False, kde_kws={"clip": (0, 500)}, hist_kws={"range": (0, 500)})
+    ax.set_yscale('log')
+    plt.xlabel('response count')
+    plt.ylabel('patches')
+    plt.title('Patch Distribution over number of responses')
+    plt.tight_layout()
+    plt.savefig('patch_dist_response_counts_500bins.pdf')
+    plt.close()
+
+    # Patch distribution by responses distinguished by with or without matching upstream
+    patch_responses_upstream_df = final[['patch_id', 'upstream', 'responses.resp_msg_id']]
+
+    patch_response_without_upstream_dask = patch_responses_upstream_df[
+        patch_responses_upstream_df['upstream'].isna() == True]
+    patch_response_without_upstream_grouped_dask = patch_response_without_upstream_dask.groupby('patch_id')[
+        'responses.resp_msg_id'].agg('count')
+
+    plt.figure(figsize=(11, 10))
+    ax = sns.distplot(patch_response_without_upstream_grouped_dask.compute(), kde=False, color='mediumvioletred', \
+                      vertical=False)
+    ax.set_yscale('log')
+    plt.title('Patch (without matching upstream) Distribution over number of responses')
+    plt.xlabel('response count')
+    plt.ylabel('patches')
+    plt.tight_layout()
+    plt.savefig('patch_without_upstream_dist_response_counts.pdf')
+    plt.close()
+
+    patch_response_with_upstream_dask = patch_responses_upstream_df[
+        patch_responses_upstream_df['upstream'].isna() == False]
+    patch_response_with_upstream_grouped_dask = patch_response_with_upstream_dask.groupby('patch_id')[
+        'responses.resp_msg_id'].agg('count')
+
+    plt.figure(figsize=(11, 10))
+    ax = sns.distplot(patch_response_with_upstream_grouped_dask.compute(), kde=False, color='olivedrab',
+                      vertical=False)
+    ax.set_yscale('log')
+    plt.title('Patch (with matching upstream) Distribution over number of responses')
+    plt.xlabel('response count')
+    plt.ylabel('patches')
+    plt.tight_layout()
+    plt.savefig('patch_with_upstream_dist_response_counts.pdf')
+    plt.close()
+
+    log.info("Saved plots, done!")
--- a/bin/pasta_prepare_evaluation.py
+++ b/bin/pasta_prepare_evaluation.py
@@ -15,8 +15,7 @@ the COPYING file in the top-level directory.
 
 from analyses import response_analysis
 from pypasta.LinuxMaintainers import LinuxMaintainers
-from pypasta import LinuxMailCharacteristics, load_linux_mail_characteristics, \
-    email_get_header_normalised, email_get_from
+from pypasta import LinuxMailCharacteristics, load_linux_mail_characteristics
 
 import argparse
 from ast import literal_eval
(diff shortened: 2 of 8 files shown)


rralf:
Hi Anmol,

can you give me an example how to start or how to run your scripts? Will leave some further comments in the commit.

Thanks!

analyses/response_analysis.py
```diff
@@ -0,0 +1,151 @@
+
```
rralf, line 1: missing copyright header

analyses/response_analysis.py
```diff
@@ -0,0 +1,151 @@
...
+
+import dask.dataframe as dd
+import matplotlib.pyplot as plt
+import pandas as pd
+import seaborn as sns
```
rralf, line 6: If you introduce new dependencies, please maintain the docker scripts. For CI purposes, PaStA needs to be able to run without any issues with the docker image.

analyses/response_analysis.py
```diff
@@ -0,0 +1,151 @@
...
+                      vertical=False)
+    ax.set_yscale('log')
+    plt.title('Patch (with matching upstream) Distribution over number of responses')
+    plt.xlabel('response count')
+    plt.ylabel('patches')
+    plt.tight_layout()
+    plt.savefig('patch_with_upstream_dist_response_counts.pdf')
+    plt.close()
+
+    log.info("Saved plots, done!")
```
rralf, line 151: Note: I didn't review the rest, I'd like to run it first.

bin/pasta_prepare_evaluation.py
```diff
@@ -13,11 +13,23 @@
...
+
 import argparse
+from ast import literal_eval
 import csv
+import dask.dataframe as dd
+import email
+import flat_table
+import numpy as np
 import os
+import pandas as pd
```
rralf, line 28: Same here, please update the docker files and test if the image is still able to run pasta.

bin/pasta_prepare_evaluation.py
```diff
@@ -26,8 +38,7 @@
 
 from tqdm import tqdm
 
-from pypasta.LinuxMaintainers import LinuxMaintainers
-from pypasta.LinuxMailCharacteristics import load_linux_mail_characteristics
+sys.path.append(os.path.join(os.path.abspath(os.pardir), 'analyses'))
```
rralf, line 40: I wonder if there's a better way to do the imports than the current sys.path.append...-hack that we do in PaStA... I need to check that.

bin/pasta_prepare_evaluation.py
```diff
@@ -343,33 +353,229 @@ def _load_responses_dict(msg_id, response_list):
...
+        except ValueError:
+            return s
+
+    def _get_message_field(msg, field):
+        if not (np.all(pd.isnull(msg))):
+            return email.message_from_bytes(msg)[field]
+        else:
+            return None
+
+    dd1 = dd.read_csv(config.f_denorm_responses, blocksize=1e9, dtype={"idx ": "int32", "patch_id ": "category",
```
rralf, line 417: So if we introduce dask + pandas as a dependency, we should use it elsewhere as well, where we can benefit. E.g., the startup of PaStA is pretty slow when loading the indexes. I'll check if we can use pandas for reading the csvs - so if we introduce pands, it'd be great if there's a chance to boost or improve other things with it

rralf: BTW, what influence does the blocksize have?

bin/pasta_prepare_evaluation.py
```diff
@@ -343,33 +353,229 @@ def _load_responses_dict(msg_id, response_list):
...
+
+    log.info("Unique response authors {}".format(final['response_author'].nunique().compute(num_workers=20)))
+
+    final.to_csv(config.f_responses_authors, single_file=True)
+
+
+def _is_response_from_bot(message_id):
+    message = _repo.mbox.get_messages(message_id)[0]
+    email = email_get_from(message)[1].lower()
+    bots = ['tip-bot2@linutronix.de', 'tipbot@zytor.com', 'noreply@ciplatform.org', 'syzbot',
```
rralf: Why didn't you use / improve the stuff that we have in LinuxMailcharacteristics.py?

bin/pasta_prepare_evaluation.py
```diff
@@ -343,33 +353,229 @@ def _load_responses_dict(msg_id, response_list):
...
                              'prepare data for ignored patch analysis \n'
                         )
 
-    parser.add_argument('--off-list',
-                        action='store_const',
-                        const='off-list',
-                        dest='mode',
+    parser.add_argument('--offlist',
+                        action='store_true',
+                        default=False,
```
rralf, line 541: Hmm, I guess we should use the same scheme that we used in Basak's script:
--mode offlist, --mode ignored, … or something like that.

bin/pasta_prepare_evaluation.py
```diff
@@ -343,33 +353,229 @@ def _load_responses_dict(msg_id, response_list):
...
-                        const='review',
-                        dest='mode',
+                        default=None,
+                        choices=['prepare', 'preprocess', 'merge', 'filter', 'analyze'],
                         help='prepare data for patch review analysis \n')
 
     analysis_option = parser.parse_args(argv)
 
-    if not analysis_option.mode:
+    if len(argv) == 0:
```
rralf, line 551: ... Then this is implicitly solved, as the parser will fail in case we didn't provide a choice.

rralf (changes requested):
Please have a look at the comments. Thanks!

bin/pasta_prepare_evaluation.py
anmolsingh20: Yes, I thought about it too. Unfortunately this is the cleanest I could make it. Reason: We have under analyses, mixture of R and python, due to which it is not a package/module. Imports are unclean when we have a mix.

bin/pasta_prepare_evaluation.py
anmolsingh20: > BTW, what influence does the blocksize have?

This is to specify how you chunk your csv

anmolsingh20: > So if we introduce dask + pandas as a dependency, we should use it elsewhere as well, where we can benefit. E.g., the startup of PaStA is pretty slow when loading the indexes. I'll check if we can use pandas for reading the csvs - so if we introduce pands, it'd be great if there's a chance to boost or improve other things with it

It is a good idea to test this.

bin/pasta_prepare_evaluation.py
anmolsingh20: I had to extract this part out, as the method _is_from_bot in LinuxMailcharacteristics is tightly coupled with the class. This prevents me from using it when a message is not a patch (e.g. a response message) and things like is_patch, or _is_next are irrelevant for a response. This is not how it should finally look, and I look forward to your thoughts how to combine this best. I have to also admit, I do not fully understand all functionality of LinuxMailcharacteristics regarding the domain.

Looking forward to your remarks here.

analyses/response_analysis.py
anmolsingh20: Sure, I will do that.

analyses/response_analysis.py
anmolsingh20: Ok

bin/pasta_prepare_evaluation.py
anmolsingh20: Sure

bin/pasta_prepare_evaluation.py
anmolsingh20: Ok, I will have a look. I made the change because I wanted to mix this with heterogeneous choices per switch. It should be possible though.

bin/pasta_prepare_evaluation.py
rralf: > > BTW, what influence does the blocksize have?
> 
> This is to specify how you chunk your csv

chunk for what? As a test, I tried to, for example, parse indexes with panda (it's indeed faster), but I didn't see any influence of the blocksize, as there's no parallelisation or sth. similar (afaict).

bin/pasta_prepare_evaluation.py
rralf: Yes, that's correct. At the moment it's tightly coupled. @ditzlike had a similar issue -- she wanted to use MailCharacteristics to assign a mail to the latest kernel version that released at that point in time. In this case, we had tight coupling, but in fact there's no need to instantiate a MailCharacteristics instance. In her case, it was possible to factor out _patch_get_version(): ref. b429ec028860fbe095cbc9f8d741abe264f58acf .

Maybe it's (somehow) possible to convert _is_from_bot() to a staticmethod.

anmolsingh20:
> Hi Anmol,
> 
> can you give me an example how to start or how to run your scripts? Will leave some further comments in the commit.
> 
> Thanks!

Sure,
./pasta prepare_evaluation --review (prepare, preprocess, merge, filter, analyse), in that order. This ensures you have the required intermediate results, and do not hold more data in memory at once.

rralf:
> > Hi Anmol,
> > can you give me an example how to start or how to run your scripts? Will leave some further comments in the commit.
> > Thanks!
> 
> Sure,
> ./pasta prepare_evaluation --review (prepare, preprocess, merge, filter, analyse), in that order. This ensures you have the required intermediate results, and do not hold more data in memory at once.

Thanks, will test it soon!

bin/pasta_prepare_evaluation.py
anmolsingh20: > > > BTW, what influence does the blocksize have?
> > 
> > 
> > This is to specify how you chunk your csv
> 
> chunk for what? As a test, I tried to, for example, parse indexes with panda (it's indeed faster), but I didn't see any influence of the blocksize, as there's no parallelisation or sth. similar (afaict).

A dask dataframe is a made up of pandas dataframes (indexed and split). Blocksize affects these partitions when csv is ingested as a dask dataframe. It has to be chosen to balance good resource usage - cpus, memory, and scheduling overhead. I played around with it when my data was gigantic. I cannot however justify this value, may be it is an artifact of some of the trials I made on some dataframe along the way. May be the defaults are good enough. We have to check this.

rralf: Ah, ok. Thanks for the explanation! Had similar issues with the parallelisation/Pool library of python: you can specify a worksize and a chunksize. The weight of those values have massive influence on memory consumption, however, one should have a look into a magic crystal ball to find sane values :-)

bin/pasta_prepare_evaluation.py
anmolsingh20: I used LinuxMailCharacteristics and improved the _is_from_bot method in two new patches. Rebased to next as well to include your changes to the class.

pypasta/LinuxMailCharacteristics.py
```diff
@@ -92,15 +91,15 @@
...
+                 'linux-serial@vger.kernel.org', 'linux-sgx@vger.kernel.org', 'linux-sh@vger.kernel.org',
                  'linux-sparse@vger.kernel.org', 'linux-spi@vger.kernel.org', 'linux-tegra@vger.kernel.org',
                  'linux-tip-commits@vger.kernel.org', 'linux-trace-devel@vger.kernel.org',
                  'linux-unionfs@vger.kernel.org', 'linux-usb@vger.kernel.org', 'linux-watchdog@vger.kernel.org',
                  'linux-wireless@vger.kernel.org', 'linux-wpan@vger.kernel.org', 'linux-xfs@vger.kernel.org',
                  'live-patching@vger.kernel.org', 'lvs-devel@vger.kernel.org', 'netdev@vger.kernel.org',
                  'netfilter-devel@vger.kernel.org', 'platform-driver-x86@vger.kernel.org',
                  'reiserfs-devel@vger.kernel.org', 'selinux@vger.kernel.org', 'sparclinux@vger.kernel.org',
-                 'stable@vger.kernel.org','target-devel@vger.kernel.org', 'util-linux@vger.kernel.org',
+                 'stable@vger.kernel.org', 'target-devel@vger.kernel.org', 'util-linux@vger.kernel.org',
```
rralf, line 102: Yep, will merge this. By the way: This list ist derived from the public inboxes. Note to self: We added new mailing lists, so this list might be incomplete. Will have to check that.

pypasta/LinuxMailCharacteristics.py
```diff
@@ -152,12 +151,12 @@ def __init__(self, c):
...
         # needs to be addressed correctly
         if (not c.mtrs_has_lists or c.mtrs_has_list_per_section) and \
-           (not c.mtrs_has_maintainers or c.mtrs_has_maintainer_per_section):
+                (not c.mtrs_has_maintainers or c.mtrs_has_maintainer_per_section):
             self.all_lists_one_mtr_per_sub = True
 
         # Metric: At least one correct list + at least one correct maintainer
         if (not c.mtrs_has_lists or c.mtrs_has_one_correct_list) and \
-           (not c.mtrs_has_maintainers or c.mtrs_has_one_correct_maintainer):
+                (not c.mtrs_has_maintainers or c.mtrs_has_one_correct_maintainer):
```
rralf, line 159: it this the "correct" indentation?

pypasta/LinuxMailCharacteristics.py
```diff
@@ -221,48 +220,35 @@ class LinuxMailCharacteristics:
...
+        bots = ['tip-bot2@linutronix.de', 'tipbot@zytor.com', 'noreply@ciplatform.org', 'syzbot',
+                'syzkaller.appspotmail.com']
+        potential_bots = ['broonie@kernel.org', 'lkp@intel.com']
         subject = email_get_header_normalised(message, 'subject')
+        uagent = email_get_header_normalised(message, 'user-agent')
+        xmailer = email_get_header_normalised(message, 'x-mailer')
 
+        if email in bots:
+            return True
+        elif email in potential_bots and \
```
rralf, line 232: No need for the elif, as we would have returned before. Will change this during merge.

pypasta/LinuxMailCharacteristics.py
```diff
@@ -221,48 +220,35 @@ class LinuxMailCharacteristics:
...
         email = self.mail_from[1].lower()
-        bots = ['broonie@kernel.org', 'lkp@intel.com']
-        potential_bot = True in [bot in email for bot in bots]
-
-        if potential_bot and \
-           email_get_header_normalised(message, 'x-patchwork-hint') == 'ignore':
-            return True
-
+        bots = ['tip-bot2@linutronix.de', 'tipbot@zytor.com', 'noreply@ciplatform.org', 'syzbot',
+                'syzkaller.appspotmail.com']
```
rralf, line 224: Ah, thanks for picking this up! We have another undetected bot left: The Hulk Robot <hulkci@huawei.com>

pypasta/LinuxMailCharacteristics.py
```diff
@@ -221,48 +220,35 @@ class LinuxMailCharacteristics:
...
             return True
-
         # Stephen Rothwell's automated emails
-        if self.is_next and 'sfr@canb.auug.org.au' in email:
+        elif self.is_next and 'sfr@canb.auug.org.au' in email:
             return True
-
-        return False
+        else:
+            return False
```
rralf, line 251: Thanks. I'll now carefully test this change with a bunch of emails.

pypasta/LinuxMailCharacteristics.py
```diff
@@ -221,48 +220,35 @@ class LinuxMailCharacteristics:
...
     def _is_from_bot(self, message):
         email = self.mail_from[1].lower()
-        bots = ['broonie@kernel.org', 'lkp@intel.com']
-        potential_bot = True in [bot in email for bot in bots]
-
-        if potential_bot and \
-           email_get_header_normalised(message, 'x-patchwork-hint') == 'ignore':
-            return True
-
+        bots = ['tip-bot2@linutronix.de', 'tipbot@zytor.com', 'noreply@ciplatform.org', 'syzbot',
```
rralf, line 223: Will change this to a set.

docker/pasta-skeleton.dockerfile
```diff
@@ -42,7 +42,10 @@ RUN apt install -y --no-install-recommends \
 RUN pip3 --no-cache-dir install \
 	dateparser \
 	flask-bootstrap \
-	flask-nav anytree
+	flask-nav anytree \
+	flat-table \
+	pandas "dask[complete]" \
+	matplotlib seaborn
```
rralf, line 48: Don't we get those libraries via Ubuntu's official repository? I would prefer that over pip for stability reasons.

pypasta/LinuxMailCharacteristics.py
rralf: ... And I'll move the constants to the top of the class.

pypasta/LinuxMailCharacteristics.py
```diff
@@ -221,48 +220,35 @@ class LinuxMailCharacteristics:
...
+                'syzkaller.appspotmail.com']
+        potential_bots = ['broonie@kernel.org', 'lkp@intel.com']
         subject = email_get_header_normalised(message, 'subject')
+        uagent = email_get_header_normalised(message, 'user-agent')
+        xmailer = email_get_header_normalised(message, 'x-mailer')
 
+        if email in bots:
```
rralf, line 230: This check is incorrect for 'syzbot', as it's not a valid email address. However, we to classify all mails with 'syzbot' in their mail adress as bots. Will fix this.

rralf:
I manually picked the "simplify _is_from_bot" patch, and added some changes and fixed the szybot-bug. Rebased your series and pushed it to wip-aggregation-analyses. Still need to review the rest...

pypasta/LinuxMailCharacteristics.py
anmolsingh20: I did reformat code in pycharm, which has default settings for python.

pypasta/LinuxMailCharacteristics.py
anmolsingh20: ok

pypasta/LinuxMailCharacteristics.py
anmolsingh20: Ok, thanks

PR #75, opened 2020-08-03
Wip aggregation analyses

anmolsingh20:
Hi Ralf,

Picking up on the review again. 

- This branch has multiple changes by now, which I plan to refine with you step by step, and merge squash commits in the end for a clean history. Please have a look at the **preprocess step only**. I will proceed further after we have brought that into shape.

- Rebased to next

Thanks
Anmol


anmolsingh20:
As discussed today, this pull request is not necessary at the moment, hence closing it. I would then keep this work on my fork, document it best, and point you to the analyses there. Due to the difference in tools used, I see the difficulty with maintenance too.

@bulwahn  : Fyi

PR #77, opened 2020-08-20
Add support for differential analyses

rsarky:
There is a scope for optimising the analysis process when it comes to differential analyses, ie. we already have some existing analyses results in PaStA and some new patches arrive for analyses.
What PaStA currently does is it assigns each of these new patches to a single element cluster, and then it runs the complete analyses again. This results in a lot of redundant comparisons. Example:

Consider the following existing state clusters of PaStA. I have indexed each cluster for illustration purposes:
```
1. 1 2 3
2. 4 5 7
3. 6 8
```
PaStA performed around 8*8 comparisons (ignoring other thresholds that PaStA has for now). For further comparisons PaStA will use the _representative_ of each cluster, let's take the first element of each cluster above to be it's representative.ie repr( 1 2 3) = 1.

Now consider that patches 9 and 10 arrive. They will be assigned to their own single element clusters, ie:
```
1. 1 2 3
2. 4 5 6
3. 6 8
4. 9
5. 10
```
In the current situation PaStA performs 5x5 comparisons (compare representative of each cluster against the other).
But we can reduce this by only comparing representatives of existing clusters with newly arriving patches as the other comparisons have already been done in the previous step. ie we reduce the comparisons to 3x2. Additionally we will also need to compare the new patches against each other a further 2x2 comparisons. Combined a total of 5x2 comparisons which is still much less than the naive way.

This can be written in a crude mathematical way as follows:
> evaluation result = new_patches X existing_patches + new_patches X new_patches  [note that existing_patches X existing_patches has already been done and it's result exists in the patch groups file]
Thus, evaluation_result = new_patches X (existing_patches + new_patches)

### Things to consider
- What if we lose the patch groups file at some point? In this case the evaluation will have to be carried out from scratch again as 
the evaluation results that have been cached does not contain all the information

### DIFF
--- a/bin/pasta_analyse.py
+++ b/bin/pasta_analyse.py
@@ -35,7 +35,6 @@ def _evaluate_patch_list_wrapper(thresholds, args):
                                 orig, cand,
                                 parallelise=False)
 
-
 def find_cherries(repo, commit_hashes, dest_list):
     """
     find_cherries() takes a list of commit hashes, a list of potential
@@ -133,6 +132,10 @@ def analyse(config, argv):
                              'compare representatives against upstream - '
                              '(default: %(default)s)')
 
+    parser.add_argument('-differential', dest='differential', action='store_true',
+                        default=False,
+                        help='Perform a differential analysis')
+
     args = parser.parse_args(argv)
 
     config.thresholds.heading = args.thres_heading
@@ -147,6 +150,9 @@ def analyse(config, argv):
     if mbox and mode == 'succ':
         log.error('Analysis mode succ is not available in mailbox mode!')
         return -1
+
+    if not mbox and args.differential:
+        log.error('Differential analysis can only be performed in mailbox mode')
 
     f_cluster, cluster = config.load_cluster(must_exist=False)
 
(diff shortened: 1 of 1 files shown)


bin/pasta_analyse.py
```diff
@@ -295,12 +307,21 @@ def fill_result(hashes, tag):
...
+                                                     representatives, new_patches,
+                                                     parallelise=True, verbose=True,
+                                                     cpu_factor=args.cpu_factor)
+        else:
+            log.info('Starting evaluation')
+            evaluation_result = evaluate_commit_list(repo, config.thresholds,
+                                                     mbox, type,
+                                                     representatives, candidates,
+                                                     parallelise=True, verbose=True,
+                                                     cpu_factor=args.cpu_factor)
```
rralf: Both calls of evaluate_commit_list are almost the same, besides we have new_patches instead of candidates in the differential case. Can't we just say:

```
if differential:
    candidates = new_patches
```

rsarky: Yes the `if else` could probably be further simplified

rsarky: Updated

PR #78, opened 2020-08-23
Add a script to test patchwork interface and add link to blogpost in README

rsarky:
Usage of the script:
> $ ./patchwork_integration_test.sh <path-to-public-inbox> <timestamp>

Only mails after the timestamp will be considered.
Results may vary due to existing indices in PaStA. Ideally this script
should be run in PaStA's docker container.

Steps performed by the script:
1. Parse the public inbox and create a mbox file.
2. Clone Patchwork
3. Build the Patchwork container
4. Create a Patchwork project, parse and dump the archive (to get
   patchwork ids)
5. Split the mbox, use the first n% (default 50) lines of mbox as
   archive for initial import.
6. Start patchwork web server
7. Run pasta sync for initial import
8. Run pasta sync for api import
9. Run pasta's analyses and form relations
10. Run replacerelations in Patchwork
11. Remove the Patchwork container

Signed-off-by: Rohit Sarkar <rohitsarkar5398@gmail.com>


tools/create_project.py
```diff
@@ -0,0 +1,26 @@
+def create_project(name):
```
rralf: missing copyright header.

tools/create_project.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+if __name__ == '__main__':
+    import django
+    django.setup()
+    from patchwork.models import Project
+    import sys
+    create_project(sys.argv[1])
+    with open('project_id', 'w+') as f:
+        f.write(str(find_project_id(sys.argv[1])))
```
rralf: BTW: What I just thought of: Why can't we use cURL to simply do API calls to patchwork to create the project? Just a thought...

rralf: And: This file is not related to PaStA, and it's not a pasta tool. Please push that file to a fork of patchwork. The integration shall then checkout patchwork and switch to the appropriate branch.

tools/pasta-setup.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
```
rralf: Please add a short copyright notice...

tools/pasta-setup.sh
```diff
@@ -0,0 +1,7 @@
...
+set -e
+
+python create_project.py $2
+./manage.py parsearchive $1 --list-id $2
+./manage.py dumparchive $2
+tar -xf *.tar
```
rralf: And move the script to patchwork. Again, this has nothing to do with PaStA, it's a patch for patchwork.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
+#! /bin/bash
```
rralf, line 1: copyright header.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+set -e
+HOME="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+PUBIN_PATH=$1
+SINCE=$2
```
rralf, line 23: I can't find a documentation of cmdline parameters. You wrote I shall run './patchwork_integration_test.sh'. That fails of course. Please document how to invoke the script.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+
+parse_pubin() {
+	git -C $PUBIN_PATH --no-pager log --pretty=format:%H --since=$SINCE --reverse | \
+		xargs -n 1 -I {} bash -c "
+				git -C $PUBIN_PATH --no-pager show --no-patch --pretty=format:'From %H %ad' {};
+				echo;
+				git -C $PUBIN_PATH --no-pager show {}:m;
+				echo -en '\n\n';
+			"
+}
```
rralf: Please outsource the script to an own tool. This tool can be useful for any other conversion of public inboxes to mboxes. Besdies that, parse_pubin does not reflect the functionality of the routine.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+	fi
+	config=$(cat <<-END
+		[PaStA]
+		MODE = "mbox"
+		UPSTREAM = "HEAD^..HEAD"
+
+		[mbox]
+		[mbox.patchwork]
+		url='http://$container_addr:8000/api/'
+		projects = [{id=$project_id, initial_archive="$initial_archive", list_email="test@test.com"}]
```
rralf: test@test.com is not the valid email address. Please pass the correct list email address via commandline.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+	if [ $1 = "mbox" ]
+	then
+		initial_archive=$INITIAL_IMPORT_MBOX
+	else
+		initial_archive=""
+	fi
+	config=$(cat <<-END
+		[PaStA]
+		MODE = "mbox"
+		UPSTREAM = "HEAD^..HEAD"
```
rralf, line 47: I don't understand this line.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+	)
+}
+
+cd "$HOME"
+
+echo "Parsing pubin..."
+parse_pubin > $MBOX
+if [ ! -d patchwork ]
+then
+	git clone https://github.com/getpatchwork/patchwork.git
```
rralf: Yep, and the next step would be to checkout the branch that contains all changes that we need.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+fi
+cp ./pasta-setup.sh patchwork/
+cp ./create_project.py patchwork/
```
rralf: ... And that wouldn't be required then.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+cp ./$MBOX patchwork/
+cd patchwork
+rm -f *.tar
+echo "UID=$UID" > .env
+echo "GID=`id -g`" >> .env
+docker-compose build
+docker-compose run --rm web ./pasta-setup.sh $MBOX $PROJECT
```
rralf, line 83: Yes, looks good!

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+
+n=$(( $(wc -l $PROJECT.mbox | cut -d" " -f 1) * $INITIAL_IMPORT_SIZE / 100 ))
```
rralf: What's n? Please leave some comments above that explain what you're doing.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+head -n $n $PROJECT.mbox > $INITIAL_IMPORT_MBOX
+mkdir -p ../../resources/linux/resources/mbox/patchwork/
+mv $INITIAL_IMPORT_MBOX $_
+
+echo Starting Patchwork web server...
+docker-compose run --rm --name pasta-patchwork web ./manage.py runserver 0.0.0.0:8000 &
+sleep 5 # Wait for container to start
```
rralf: dangerous... What if the container isn't ready after 5s? Please wait for the availability of the container (e.g., dummy API call via cURL or sth like that).

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+docker-compose build
+docker-compose run --rm web ./pasta-setup.sh $MBOX $PROJECT
+
+n=$(( $(wc -l $PROJECT.mbox | cut -d" " -f 1) * $INITIAL_IMPORT_SIZE / 100 ))
+head -n $n $PROJECT.mbox > $INITIAL_IMPORT_MBOX
+mkdir -p ../../resources/linux/resources/mbox/patchwork/
+mv $INITIAL_IMPORT_MBOX $_
+
+echo Starting Patchwork web server...
+docker-compose run --rm --name pasta-patchwork web ./manage.py runserver 0.0.0.0:8000 &
```
rralf: Doesn't docker-compose have an option to start things as daemon/background mode? Just detaching the container to background is dangerous.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+sleep 5 # Wait for container to start
+container_addr=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' pasta-patchwork)
```
rralf: Yikes... Which means that PaStA's config file depends on the exact IP-address of the container. I think it should be possible to expose the whole patchwork instance to the host with docker-compose ports: https://docs.docker.com/compose/compose-file/#ports

If we do so, we would ensure that we can access 127.0.0.1:8000 from the host, and wouldn't have to pin the IP address of the container.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+
+project_id=$(cat project_id)
+rm project_id
```
rralf: What happens here? Please spend a few words on why/what you do here.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+# api import
+./pasta sync -mbox
+
+./pasta analyse rep
+./pasta rate
+./pasta form_patchwork_relations
+
+mv resources/linux/resources/patch-groups-patchwork tools/patchwork/
+cd $HOME/patchwork
+docker-compose run --rm web ./manage.py replacerelations patch-groups-patchwork
```
rralf, line 125: Yep, the rest looks good so far.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,102 @@
...
+
+
```
rralf: trailing newlines

rralf (changes requested):
Hi Rohit,

thanks for the patch! But I think we'll need another round or two. Anyway, it's going into the correct direction, but we need more structure. Please add some more (short) comments, on what you're trying to achieve. For example, l.84 or l.91 are really helpful for readers to understand what you're currently doing in your script.

Besides that - please move the whole patchwork-stuff over to patchwork, and checkout your branch. Please find the rest of the comments inline.

Thanks
  Ralf

tools/create_project.py
rsarky: > BTW: What I just thought of: Why can't we use cURL to simply do API calls to patchwork to create the project? Just a thought...

There is no API in Patchwork to create projects as of now

tools/patchwork_integration_test.sh
rsarky: Yeah I think it is called the detached mode, I can use that.
Just curious, why is detaching to the background more dangerous than running the container in detached mode?

EDIT:
It is certainly beneficial to use the detached mode in our case. Because the CLI blocks until the docker command returns then

tools/patchwork_integration_test.sh
rsarky: How can we be sure of `127.0.0.1`. What if the script is not run from inside the container? In hydra the container addresses turn out to be of the form 192.168.6.x

tools/patchwork_integration_test.sh
rsarky: I documented it in the commit message. Will add some usage documentation inside the script too

tools/patchwork_integration_test.sh
rsarky: That address doesn't really play a role in the analyses. I think we dont need to pass it from the command line

tools/patchwork_integration_test.sh
rsarky: ~Yeah this is certainly hacky, okay I will try to poll the container using cURL~

EDIT:
This is solved using the detached mode:
In this case the docker command returns the name of the container when the container is started, hence the cli is blocked until the command returns

tools/patchwork_integration_test.sh
rsarky: Agreed, moved to a file named `pubin2mbox.sh`

tools/create_project.py
rralf: Sure, there is. Otherwise you wouldn't be able to create a project with the web interface. You know the request that will be submitted when you create a project. That can also be called via curl.

tools/patchwork_integration_test.sh
rralf: Please explain your setup: Where do you run PaStA (the host?) and where do you run patchwork (the container?). If so, then pasta could directly access the API via localhost:8000

rralf: In the commit message, you wrote "ideally it should be run in the container". What if not? Will it fail?

If PaStA runs inside the container, would it be possible to somehow forward the host port where we can access the API? In this case, we could also just access :8000 inside the container.

tools/patchwork_integration_test.sh
rralf: Still, test@test.com is incorrect.

tools/create_project.py
rsarky: I mean there is no publically exposed API for that. That probably uses some internal API. Reference: https://patchwork.readthedocs.io/en/latest/api/rest/schemas/v1.2/

tools/patchwork_integration_test.sh
rsarky: The reason for me saying "ideally it should be run in the container" is that in that case there is no existing index files or mboxes that can mess with the test results. For eg, if the public inbox you are trying to ingest through the script is already imported in pasta the script wont really do much

rsarky: > Please explain your setup: Where do you run PaStA (the host?) and where do you run patchwork (the container?). If so, then pasta could directly access the API via localhost:8000

You can run this script anywhere, doesnt matter whether it's in the container or not. Hence I get the ip address of the container. I dont see a problem with this. What is your concern?

tools/create_project.py
rralf: POST request on http://localhost:8000/admin/patchwork/project/add/, linkname: foo, listemail: a@b.com, ...

rralf: Would be a single line in the shell script. Though I don't know how complex it is to forward cookies, ... So that's a thing with really low priority.

tools/patchwork_integration_test.sh
rralf: I'd prefer if the config file would be independant of the IP address. So I think if there's some other methodology to announce the service to the PaStA Container / Host. Ideally, the config just holds localhost:8000. In case PaStA runs on the host, then it'll work out of the box. If PaStA runs inside the container, then we won't see patchwork via localhost:8000. Could you try if that would be possible with the docker option --net=host?

tools/create_project.py
rsarky: Hmm yeah. Anyway doing it through python is reliable enough, because I am assuming if we use the admin API we will have to create an admin account too. And then somehow get a cookie to send along with the request

rralf: Yes true, so this is really low prio.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,113 @@
...
+# the COPYING file in the top-level directory.
+
+set -e
+USAGE=$(cat <<-END
+	$ ./patchwork_integration_test.sh <path-to-public-inbox> <timestamp>
+	Only mails in public inbox since the given <timestamp> will be considered
+	END
+	)
+if [ $# -eq 0 ]
+then
```
rralf, line 16: incorrect.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,113 @@
...
+	echo -e "$USAGE"
+	exit 1
+fi
+
+HOME="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
rralf: HOME is a reserved environment variable.

tools/patchwork_integration_test.sh
rsarky: > I'd prefer if the config file would be independant of the IP address

That's possible.

>  If PaStA runs inside the container, then we won't see patchwork via localhost:8000

Not sure about this. Maybe it can work in this case too. So want to test it first.
Was trying to build the pasta docker images but there is some issue, are you able to build the images?

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,113 @@
...
+if [ ! -d patchwork ]
+then
+	git clone https://github.com/rsarky/patchwork.git
+fi
+
+cp ./$MBOX patchwork/
+
+cd patchwork
+git checkout pasta-patchwork-integration
+rm -f *.tar
```
rralf, line 70: What are you deleting here?

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,113 @@
...
+echo "UID=$UID" > .env
+echo "GID=`id -g`" >> .env
+docker-compose build
+docker-compose run --rm web ./pasta-setup.sh $MBOX $PROJECT
```
rralf, line 83: Please leave a comment what this comment is meant to do. E.g.: Add $MBOX to patchwork (parsearchive), and export it again so that the result contains x-patch-id hdrs.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,113 @@
...
+
+# the pasta-setup script generates a file "project_id" that contains the
+# project id of the project created.
+project_id=$(cat project_id)
+rm project_id
+
+# Get the number of lines in the first INITIAL_IMPORT_SIZE % of the mbox
+num_lines=$(( $(wc -l $PROJECT.mbox | cut -d" " -f 1) * $INITIAL_IMPORT_SIZE / 100 ))
```
rralf, line 91: Ah, with the comment I understand what you're doing here. Did you check if you can solve that with the split command?

tools/pubin2mbox.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash
```
rralf, line 1: Copyright :)

rralf (changes requested):
Find the comments inline.

tools/patchwork_integration_test.sh
rralf: What are you deleting here?

README.md
```diff
@@ -261,6 +261,11 @@ PaStA and Patchwork
 --------------------
 The results of PaStA's analyses can be used by [Patchwork](http://jk.ozlabs.org/projects/patchwork/).
 
+Work on PaStA and Patchwork integration was also undertaken as a [Google Summer
+of Code](https://summerofcode.withgoogle.com/) project in 2020. The work done
+is described in detail in this [blog
+post](https://rsarky.github.io/2020/08/10/pasta-patchwork.html)
```
rralf: Could you please add the raw content to PaStA? I don't want to loose those documents at some point.

tools/patchwork_integration_test.sh
rralf: The last commit on docker/ was on Aug 5. Back then it worked. What's the issue?

tools/patchwork_integration_test.sh
rsarky: Umm, so do we take the list address as an argument from the user? That seems redundant

README.md
rsarky: Yes, should I create a  `documentation` directory and add it there?

tools/patchwork_integration_test.sh
rsarky: I didnt get you? Why is this incorrect?

tools/patchwork_integration_test.sh
rsarky: Upon using dumparchive, patchwork dumps a tar archive of the mbox. The name of the tar archive is `patchwork-<timestamp>.tar`. I am deleting any existing archives from previous invocations of dumparchive.

tools/patchwork_integration_test.sh
rsarky: `docker-skeleton` doesnt seem to be building. Issues with the ubuntu repository urls I think. Not sure

tools/patchwork_integration_test.sh
rsarky: Yes. Even using the split command we need to find the exact number of lines that constitute a certain proportion of the file

tools/patchwork_integration_test.sh
rralf: I'd suppose that. why would that be redundant? On patchwork-side we also have to specify a project name AND a list email address, which seems kind of redundant as well.

README.md
rralf: Sounds good!

tools/patchwork_integration_test.sh
rralf: If $# = 1, you won't print the usage and continue, while the command line is not correct.

tools/patchwork_integration_test.sh
rralf: AFAICT, you extract the archive from the script that runs in the container. We don't we directly remove the artefact there?

tools/patchwork_integration_test.sh
rsarky: This was deliberate, I am only printing usage when the script is called without parameters. Instead I can do a `$# -ne 2` perhaps

tools/patchwork_integration_test.sh
rsarky: Sure can do that too

tools/patchwork_integration_test.sh
rralf: Just did a test build - I can't see an issue.

rsarky: That's weird. I am on Hydra
The command I am using is `docker build -f docker/pasta-skeleton.dockerfile .`

tools/patchwork_integration_test.sh
rsarky: Ok, added another argument.

README.md
rsarky: done

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,115 @@
...
+if [ $# -ne 2 ]
+then
+	echo -e "$USAGE"
+	exit 1
+fi
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+PUBIN_PATH=$1
+SINCE=$2
+MBOX=test-mbox.mbox
```
rralf, line 25: Can we store that under /tmp/$listname.mbox? No need to mess up tools/ with mbox files

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,115 @@
...
+
+get_config() {
+	# Get a minimal config for PaStA
+	# For api imports we keep the intial archive field empty
+	if [ $1 = "mbox" ]
+	then
+		initial_archive=$INITIAL_IMPORT_MBOX
+	else
+		initial_archive=""
+	fi
```
rralf, line 43: initial_archive = ""
if [$1 = "mbox"]; then
    initial_archive=$INITIAL...
fi

is shorter.

tools/patchwork_integration_test.sh
rralf: It's still test@test.com

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,115 @@
...
+		[mbox.patchwork]
+		url='http://localhost:8000/api/'
+		projects = [{id=$project_id, initial_archive="$initial_archive", list_email="test@test.com"}]
+		END
+	)
+}
+
+cd "$SCRIPT_DIR"
+
+echo "Parsing pubin..."
```
rralf: echo "Converting public inbox to mbox..."

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,115 @@
...
+docker-compose run -d --name pasta-patchwork -p 8000:8000 web ./manage.py runserver 0.0.0.0:8000
+
+cd ../..
+
+get_config mbox
+echo "$config" > resources/linux/config
+./pasta set_config linux
+
+#initial import
+./pasta sync -mbox
```
rralf: -create downstream. No need to create caches for upstream. We don't use them at the moment.

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,115 @@
...
+
+get_config api
+echo "$config" > resources/linux/config
+./pasta set_config linux
+
+# api import
+./pasta sync -mbox
```
rralf: same here.

README.md
rralf: Please link do documentation/ instead of relying on rsarky.github.io

tools/patchwork_integration_test.sh
```diff
@@ -0,0 +1,115 @@
...
+	if [ $1 = "mbox" ]
+	then
+		initial_archive=$INITIAL_IMPORT_MBOX
+	else
+		initial_archive=""
+	fi
+	config=$(cat <<-END
+		[PaStA]
+		MODE = "mbox"
+		UPSTREAM = "HEAD^..HEAD"
```
rralf, line 47: We need sane default thresholds here, otherwise we'll run into interactive rating. Please set
INTERACTIVE_THRESHOLD = 0.82
AUTOACCEPT_THRESHOLD = 0.82
DIFF_LINES_RATIO = 0.7

HEADING_THRESHOLD = 0.6
FILENAME_THRESHOLD = 0.95

rralf (changes requested):
Ok, looks much better now! I got it running. I tried to work on alsa-devel with 2020-01-01 as time window, and it worked. But I saw another issue: There were ~30 patches in the invalid file. Interestingly I found entries for those files in the index, but I couldn't find their raw content under resources/linux/resources/mbox/patchwork/y/m/d/hash. Somehow those files were referenced but missing. Will have to rerun the script to see what happens. Will get back to you soon. Thanks!

rralf:
Ah... And in your integration script: Please ensure that you remove preexisting caches before with ./pasta sync -noup -clear all. Secondly, please remove all invalids with rm resources/linux/resources/mbox/invalid/* to get a clean working environment. Thanks!

rralf:
>  Interestingly I found entries for those files in the index, but I couldn't find their raw content under resources/linux/resources/mbox/patchwork/y/m/d/hash. Somehow those files were referenced but missing. Will have to rerun the script to see what happens. Will get back to you soon. 

Ok, that was my fault, I mixed up the patchwork index with the pubin index.

rsarky:
> Ah... And in your integration script: Please ensure that you remove preexisting caches before with ./pasta sync -noup -clear all. Secondly, please remove all invalids with rm resources/linux/resources/mbox/invalid/* to get a clean working environment. Thanks!

Sure! This way we get repeatable results from the script

rralf:
Thanks, applied to next with some changes.

PR #81, opened 2021-07-14
docker: Add optional dependency Levenshtein

8sd:
After installing the listed requirements pasta said:

```
UserWarning: Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning
```

This PR fixes this.


rralf:
Thanks, applied to next. I also added a patch that forwards Ubuntu 20.04 to 21.04.

PR #82, opened 2021-08-03
Extend mail characteristics2

8sd:
This PR adds Jailhouse-support for the ignored patch analysis

### DIFF
--- a/pypasta/JailhouseMailCharacteristics.py
+++ b/pypasta/JailhouseMailCharacteristics.py
@@ -1,0 +1,63 @@
+"""
+PaStA - Patch Stack Analysis
+
+Copyright (c) OTH Regensburg, 2021
+
+Author:
+  Sebastian Duda <git@sebdu.de>
+
+This work is licensed under the terms of the GNU GPL, version 2.  See
+the COPYING file in the top-level directory.
+"""
+
+from .MailCharacteristics import MailCharacteristics, PatchType
+
+
+class JailhouseMailCharacteristics(MailCharacteristics):
+    ROOT_DIRS = ['ci',
+                 'configs',
+                 'Documentation',
+                 'driver',
+                 'hypervisor',
+                 'include',
+                 'inmates',
+                 'pyjailhouse',
+                 'scripts',
+                 'tools',
+    ]
+    ROOT_FILES = ['CONTRIBUTING.md',
+                  'COPYING',
+                  'FAQ.md',
+                  '.git',
+                  '.gitignore',
+                  'Kbuild',
+                  'LICENSING.md',
+                  'Makefile',
+                  'README.md',
+                  'setup.py',
+                  'TODO.md',
+                  '.travis.yml',
+                  'VERSION',
+    ]
+
+    # Additional lists that are not known by pasta
+    LISTS = set()
+
+    HAS_MAINTAINERS = False
+
+    def __init__(self, repo, maintainers_version, clustering, message_id):
+        super().__init__(repo, clustering, message_id)
+        self.__init(repo, maintainers_version)
+        self._cleanup()
+
+    def __init(self, repo, maintainers_version):
+        if self.is_from_bot:
+            self.type = PatchType.BOT
+
+        if not self.is_patch:
+            return
+
+        if self.type == PatchType.OTHER:
+            self.type = PatchType.PATCH
+
+        self._integrated_correct(repo, maintainers_version)
--- a/pypasta/MailCharacteristics.py
+++ b/pypasta/MailCharacteristics.py
@@ -174,8 +174,15 @@ class MailCharacteristics:
         return False
 
     def _integrated_correct(self, repo, maintainers_version):
+        if self.first_upstream in repo:
+            upstream = repo[self.first_upstream]
+            self.committer = upstream.committer.name.lower()
+
         if maintainers_version is None:
             return
+
+        # stuff for maintainers analysis
+        self.maintainers = dict()
 
         maintainers = maintainers_version[self.version]
         sections = maintainers.get_sections_by_files(self.patch.diff.affected)
@@ -197,8 +204,6 @@ class MailCharacteristics:
         # integrated by a maintainer that is responsible for a section that is
         # affected by the patch. IOW: The field indicates if the patch was
         # picked by the "correct" maintainer
-        upstream = repo[self.first_upstream]
-        self.committer = upstream.committer.name.lower()
         self.integrated_correct = False
         self.integrated_xcorrect = False
         sections = maintainers.get_sections_by_files(upstream.diff.affected)
@@ -236,6 +241,9 @@ class MailCharacteristics:
                 break
 
     def list_matches_patch(self, list):
+        if not self.maintainers:
+            return None
+
         for lists, _, _ in self.maintainers.values():
             if list in lists:
                 return True
@@ -262,7 +270,7 @@ class MailCharacteristics:
         self.lists = repo.mbox.get_lists(message_id)
 
         # stuff for maintainers analysis
-        self.maintainers = dict()
+        self.maintainers = None
 
         # Patch characteristics
         self.is_patch = message_id in repo and message_id not in repo.mbox.invalid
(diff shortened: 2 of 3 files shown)


pypasta/MailCharacteristics.py
```diff
@@ -171,12 +171,31 @@ def _is_from_bot(self):
...
+        # Buildroot's daily results bot
+        if 'oe-core cve metrics' in subject.lower():
+            return True
+
         return False
 
     def _integrated_correct(self, repo, maintainers_version):
+        if self.first_upstream in repo:
+            upstream = repo[self.first_upstream]
+            self.committer = upstream.committer.name.lower()
```
8sd, line 191: Wollen wir hier vielleicht die Mailadresse ausleiten? Die ist aussagekräftiger, oder?

pypasta/JailhouseMailCharacteristics.py
```diff
@@ -0,0 +1,63 @@
...
+
+This work is licensed under the terms of the GNU GPL, version 2.  See
+the COPYING file in the top-level directory.
+"""
+
+from .MailCharacteristics import MailCharacteristics, PatchType
+
+
+class JailhouseMailCharacteristics(MailCharacteristics):
+    ROOT_DIRS = ['ci',
```
8sd, line 17: Hier müsstest du mir noch das Skript zukommen lassen, mit dem du alle Versionen überprüft hast. 
Andernfalls scripte ich das selbst.

pypasta/JailhouseMailCharacteristics.py
```diff
@@ -0,0 +1,63 @@
+"""
+PaStA - Patch Stack Analysis
+
+Copyright (c) OTH Regensburg, 2021
```
rralf, line 4: OTH is not the copyright holder in this case. So this is either your university, or you.

pypasta/JailhouseMailCharacteristics.py
rralf: Yep, just sent it to you. The directories seem complete!

pypasta/JailhouseMailCharacteristics.py
```diff
@@ -0,0 +1,63 @@
...
+                  '.git',
+                  '.gitignore',
+                  'Kbuild',
+                  'LICENSING.md',
+                  'Makefile',
+                  'README.md',
+                  'setup.py',
+                  'TODO.md',
+                  '.travis.yml',
+                  'VERSION',
```
rralf, line 40: We miss:

driver.c
jailhouse.h
README (there was an ancient version without .md)
TODO (same here)

Rest looks good!

pypasta/MailCharacteristics.py
rralf: Okay, I remember that - for some reason - we talked about this change, but I don't understand at the moment why we did this. Could you please factor this change out to an own commit with an explanation why we do this? This change is preparatory work for supporting other projects, and not related to the introduction of Jailhouse.

pypasta/MailCharacteristics.py
```diff
@@ -197,8 +204,6 @@ def check_maintainer(section, committer):
         # integrated by a maintainer that is responsible for a section that is
         # affected by the patch. IOW: The field indicates if the patch was
         # picked by the "correct" maintainer
-        upstream = repo[self.first_upstream]
-        self.committer = upstream.committer.name.lower()
```
rralf, line 201: And what about the above comments? They now point into the void?

pypasta/MailCharacteristics.py
```diff
@@ -327,11 +335,13 @@ def load_maintainers_characteristics(config, characteristics_class, clustering,
...
-    patches = ids - repo.mbox.invalid
-    tags = {repo.patch_get_version(repo[x]) for x in patches}
-    maintainers_version = load_maintainers(config, tags)
+    maintainers_version = None
+    if characteristics_class.HAS_MAINTAINERS:
+        # We can safely limit to patches only, as only patches will be used for the
+        # maintainers analysis.
+        patches = ids - repo.mbox.invalid
+        tags = {repo.patch_get_version(repo[x]) for x in patches}
+        maintainers_version = load_maintainers(config, tags)
```
rralf, line 356: What about all other projects that do not have HAS_MAINTAINERS yet? They are now broken with this change. Please factor this change out to a preparatory patch, then we get a better overview how we can handle this.

pypasta/MailCharacteristics.py
```diff
@@ -373,11 +383,13 @@ def load_characteristics(config, clustering, message_ids = None):
     config.mbox_timewindow, and loads multiple instances of maintainers for the
     patches of the clustering.
     """
+    from .JailhouseMailCharacteristics import JailhouseMailCharacteristics
     from .LinuxMailCharacteristics import LinuxMailCharacteristics
     from .QemuMailCharacteristics import QemuMailCharacteristics
     from .UBootMailCharacteristics import UBootMailCharacteristics
     from .XenMailCharacteristics import XenMailCharacteristics
     _load_characteristics = {
+        'jailhouse': (load_maintainers_characteristics, JailhouseMailCharacteristics),
```
rralf, line 404: And if we once have factored out those things, we might even see a better way to handle this dictionary here.

pypasta/Repository/Repository.py
```diff
@@ -31,6 +31,7 @@
 _tmp_repo = None
 
 mainline_regex = {
+    'jailhouse': re.compile(r'^v.*$'),
```
rralf, line 34: Looks good.

rralf (changes requested):
see inline comments

pypasta/MailCharacteristics.py
```diff
@@ -76,7 +76,7 @@ class MailCharacteristics:
     BOTS = {'tip-bot2@linutronix.de', 'tipbot@zytor.com',
             'noreply@ciplatform.org', 'patchwork@emeril.freedesktop.org'}
     POTENTIAL_BOTS = {'broonie@kernel.org', 'lkp@intel.com'}
-    PROCESSES = ['linux-next', 'git pull', 'rfc']
+    PROCESSES = ['linux-next', 'git pull', 'rfc', '[PULL]']
```
rralf, line 79: Yeah, sounds valid. But what about having everything here in lower case, and comparing it against lower case? With this, we would also catch [pull].

pypasta/MailCharacteristics.py
```diff
@@ -171,6 +171,18 @@ def _is_from_bot(self):
         if self.is_next and 'sfr@canb.auug.org.au' in email:
             return True
 
+        # Github Bot
+        if 'noreply@github.com' in email:
+            return True
+
+        # Buildroot's daily results bot
+        if '[autobuild.buildroot.net] Daily results' in subject:
```
rralf, line 179: Please compare against subject.lower()

pypasta/MailCharacteristics.py
```diff
@@ -171,6 +171,18 @@ def _is_from_bot(self):
...
+            return True
+
+        # Buildroot's daily results bot
+        if 'oe-core cve metrics' in subject.lower():
```
rralf, line 183: And then you can combine those two cases.

rralf: Besides that, looks good!

rralf (changes requested):

pypasta/MailCharacteristics.py
8sd: We need this to get the committer of patches in projects without a maintainer.
Without this change, we would return from this function before we get this information.
The information about the committer however is relevant and valid for all projects.

pypasta/MailCharacteristics.py
8sd: this is not necessary since `subject = email_get_header_normalised(self.message, 'subject')` already lowers the subject.

rralf: Oh, you're right - but then 'Daily results' will never match.

pypasta/MailCharacteristics.py
rralf: And the .lower() is redundant here.

